PR #1, opened 2020-04-14
0.0

scareything:
First cut at the tunneler SDK. Not complete yet but far enough along to start integrating if/when ready.

- A prototype tunneler app is in programs/ziti-tunneler. This will likely be moved into its own repository for the linux/darwin CLI tunneler once the SDK is farther along.
- The netif_driver API should allow us to plug a wintun "driver" into lwip.
- No DNS yet, so your service configs will need to be defined with IP addresses. And you'll need to manually add routes to your network interface.


README.md
```diff
@@ -1 +1,43 @@
...
+
+
+
+- `to_ziti(uint8_t *packet, size_t len)`
+
+  Implementing applications must call this function when a packet is
+  read from the virtual NIC.
+
+
+- `to_client(uint8_t *packet, size_t len)`
```
smilindave26, line 29: is this the same as to_client_cb?

deps/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+project(ziti-tunneler-sdk-deps)
+
+include(ExternalProject)
+
+add_subdirectory(ziti-sdk-c)
```
smilindave26, line 5: no newline at end of file.   lwip and lwip-contrib?

lib/ziti_utils.c
```diff
@@ -0,0 +1,207 @@
...
+ * https://sourceforge.net/p/predef/wiki/OperatingSystems/
+ */
+#if defined(WIN32)
+#define ZITI_OS Windows
+#elif defined(__ANDROID__)
+#define ZITI_OS Android
+#elif defined(__linux__)
+#define ZITI_OS Linux
+#elif defined(__APPLE__)
+#define ZITI_OS MacOS
```
smilindave26, line 50: iOS eventually?

scareything: Not sure. This file was copied from the ziti SDK for the logger. We're going to surface it in the ziti SDK though.

deps/CMakeLists.txt
scareything: lwip and lwip-contrib are added in a more indirect way - by including lwip/src/Filelists.cmake.

README.md
scareything: This entire readme is stale, will update when the dust settles.

PR #3, opened 2020-04-16
Win32 tun build

ekoby:


lib/lwip/lwipopts.h
```diff
@@ -27,6 +27,9 @@
 #define LWIP_NETCONN 0
 #define LWIP_SOCKET 0
 
+#ifdef _WIN32
+#define LWIP_NORAND 1
```
dovholuknf, line 31: was this for some specific purpose? might be nice to leave a comment behind as to 'why'?

PR #6, opened 2020-04-20
Extract tunneler callbacks

scareything:
Move the tunneler callbacks into a lib so other ziti-sdk-c apps can re-use them.


include/nf/ziti_tunneler_cbs.h
```diff
@@ -0,0 +1,36 @@
...
+
+/** context passed through the tunneler SDK for network i/o */
+typedef struct ziti_io_ctx_s {
+    ziti_conn_state      state;
+    nf_connection        nf_conn;
+    tunneler_io_context  tnlr_io_ctx;
+} ziti_io_context;
+
+/** called by tunneler SDK after a client connection is intercepted */
+void *my_ziti_dial(const char *service_name, const void *ziti_ctx, tunneler_io_context tnlr_io_ctx);
```
ekoby, line 24: I'd remove `my_` prefixes

lib/CMakeLists.txt
```diff
@@ -37,8 +47,8 @@ target_include_directories(ziti_tunneler
         PRIVATE ../include/priv
         )
 
-add_library(lwipwin32arch STATIC
-        ${LWIP_CONTRIB_DIR}/ports/win32/sys_arch.c)
+#add_library(lwipwin32arch STATIC
```
ekoby, line 50: this needs to be wrapped in `if (WIN32)`

scareything: Ahh, I meant to un-comment this back in, as I made it conditional in another branch.

PR #27, opened 2020-05-26
Update submodule to https vs git

dovholuknf:
travis has a hard time with git submodules - moving to https instead


scareything:
do you understand what is going on in travis? changing to https will be a bit of a bummer when touching ziti-sdk-c in conjunction with the tunneler sdk.

PR #33, opened 2020-07-09
Lwip stability

scareything:
This PR mainly fixes an issue that caused the tunneler SDK to incorrectly intercept retried SYN segments instead of letting lwip handle them (commit 1c252ad). With this change the Darwin and linux tunnelers have been running iperf loops for hours.

Along the way I also added some logging improvements that are worthwhile. 

Finally, this PR includes a change to the Darwin tun polling that reduces latency by letting uv poll the tun directly (via uv_poll) instead of calling `poll()` from a uv_check callback (commit 1214626).

### DIFF
--- a/lib/tunneler_tcp.c
+++ b/lib/tunneler_tcp.c
@@ -82,11 +82,17 @@ static err_t on_tcp_client_data(void *io_ctx, struct tcp_pcb *pcb, struct pbuf *
         ZITI_LOG(INFO, "conn was closed err=%d", err);
         return ERR_OK;
     }
-    ZITI_LOG(DEBUG, "on_tcp_client_data status %d", err);
+    ZITI_LOG(VERBOSE, "status %d", err);
     struct io_ctx_s *_io_ctx = (struct io_ctx_s *)io_ctx;
     tunneler_io_context tnlr_io_ctx = *_io_ctx->tnlr_io_ctx_p;
 
+    if (tnlr_io_ctx == NULL) {
+        ZITI_LOG(INFO, "null tnlr_io_ctx");
+        return ERR_CONN;
+    }
+
     if (err == ERR_OK && p == NULL) {
+        ZITI_LOG(INFO, "client closed connection: client=%s, service=%s", tnlr_io_ctx->client, tnlr_io_ctx->service_name);
         tcp_close(pcb);
         tnlr_io_ctx->tnlr_ctx->opts.ziti_close(_io_ctx->ziti_io_ctx);
         _io_ctx->ziti_io_ctx = NULL;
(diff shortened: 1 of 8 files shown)


lib/tunneler_tcp.c
```diff
@@ -236,7 +261,7 @@ u8_t recv_tcp(void *tnlr_ctx_arg, struct raw_pcb *pcb, struct pbuf *p, const ip_
     u16_t src_p = lwip_ntohs(tcphdr->src);
     u16_t dst_p = lwip_ntohs(tcphdr->dest);
 
-    ZITI_LOG(DEBUG, "received segment %s:%d->%s:%d",
+    ZITI_LOG(TRACE, "received segment %s:%d->%s:%d",
```
dovholuknf, line 264: maybe this should be behind an compiler directive? for 'hot path' stuff like this even this is work that we might not want/need?

lib/ziti_tunneler_priv.h
```diff
@@ -21,6 +21,7 @@ typedef enum  {
 struct tunneler_io_ctx_s {
     tunneler_context    tnlr_ctx;
     const char *        service_name;
+    char                client[32];
```
dovholuknf, line 24: 32 chars will be enough? :)

lib/tunneler_tcp.c
```diff
@@ -82,11 +82,17 @@ static err_t on_tcp_client_data(void *io_ctx, struct tcp_pcb *pcb, struct pbuf *
         ZITI_LOG(INFO, "conn was closed err=%d", err);
         return ERR_OK;
     }
-    ZITI_LOG(DEBUG, "on_tcp_client_data status %d", err);
+    ZITI_LOG(VERBOSE, "status %d", err);
     struct io_ctx_s *_io_ctx = (struct io_ctx_s *)io_ctx;
     tunneler_io_context tnlr_io_ctx = *_io_ctx->tnlr_io_ctx_p;
 
+    if (tnlr_io_ctx == NULL) {
+        ZITI_LOG(INFO, "null tnlr_io_ctx");
```
dovholuknf, line 90: this isn't an ERROR?

lib/tunneler_tcp.c
scareything: Are you suggesting wrapping this and probably other logging statements in an #ifdef DEBUG or something similar? If so, that would prevent ordinary folk (tunneler app users) from increasing the log level unless they have a debug build of the app in their possession. If this is the idea then I'd rather first determine that the performance hit is worth the reduced supportability.

lib/ziti_tunneler_priv.h
scareything: 32 is plenty. e.g.:

    tcp:123.456.789.012:123456

and it is populated with snprintf.

lib/tunneler_tcp.c
scareything: not necessarily. the connection could have been closed from the "ziti" side of the tunnel, which would null out the tunneler i/o context.

PR #38, opened 2020-08-03
Dns support dnsmasq

ekoby:
adds DNS assignment into tunneler core
adds support for dnsmasq DNS resolution

### DIFF
--- a/include/ziti/ziti_tunneler.h
+++ b/include/ziti/ziti_tunneler.h
@@ -59,9 +59,18 @@ typedef struct tunneler_sdk_options_s {
     ziti_sdk_dial_cb   ziti_dial;
     ziti_sdk_close_cb  ziti_close;
     ziti_sdk_write_cb  ziti_write;
+    const char *ip_range;
 } tunneler_sdk_options;
 
+typedef struct dns_manager_s dns_manager;
+struct dns_manager_s {
+    const char* (*map_to_ip)(dns_manager *dns, const char *host);
+    void *data;
+};
+
 extern tunneler_context ziti_tunneler_init(tunneler_sdk_options *opts, uv_loop_t *loop);
+
+extern void ziti_tunneler_set_dns(tunneler_context tnlr_ctx, dns_manager *dns);
 
 extern int ziti_tunneler_intercept_v1(tunneler_context tnlr_ctx, const void *ziti_ctx, const char *service_id, const char *service_name, const char *hostname, int port);
 
--- a/include/ziti/ziti_tunneler.h
+++ b/include/ziti/ziti_tunneler.h
@@ -64,7 +64,7 @@ typedef struct tunneler_sdk_options_s {
 
 typedef struct dns_manager_s dns_manager;
 struct dns_manager_s {
-    const char* (*map_to_ip)(dns_manager *dns, const char *host);
+    int (*apply)(dns_manager *dns, const char *host, const char *ip);
     void *data;
 };
 
--- a/include/ziti/ziti_tunneler.h
+++ b/include/ziti/ziti_tunneler.h
@@ -87,6 +87,8 @@ extern int ziti_tunneler_close(tunneler_io_context *tnlr_io_ctx);
 
 extern const char* ziti_tunneler_version();
 
+extern void ziti_tunneler_init_dns(uint32_t mask, int bits);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/ziti/ziti_tunneler.h
+++ b/include/ziti/ziti_tunneler.h
@@ -59,7 +59,6 @@ typedef struct tunneler_sdk_options_s {
     ziti_sdk_dial_cb   ziti_dial;
     ziti_sdk_close_cb  ziti_close;
     ziti_sdk_write_cb  ziti_write;
-    const char *ip_range;
 } tunneler_sdk_options;
 
 typedef struct dns_manager_s dns_manager;
--- a/lib/dns.c
+++ b/lib/dns.c
@@ -45,10 +45,10 @@ static cache ip_cache = {
         .entries = {0},
 };
 
-void init_dns(uint32_t mask, int mask_bits) {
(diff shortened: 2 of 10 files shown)


include/ziti/ziti_tunneler.h
```diff
@@ -59,10 +59,19 @@ typedef struct tunneler_sdk_options_s {
     ziti_sdk_dial_cb   ziti_dial;
     ziti_sdk_close_cb  ziti_close;
     ziti_sdk_write_cb  ziti_write;
+    const char *ip_range;
```
scareything, line 62: Would this be better named "autoip_range" or something to indicate that it's used for hostname-addressed services? Mainly I think "ip_range" might be confusing when we support network services, and also this field won't be used at all for IP-addressed services.

lib/dns.c
```diff
@@ -0,0 +1,71 @@
...
+#include <uv_mbed/queue.h>
+
+
+#define MAX_DNS_NAME 256
+#define MAX_IP_LENGTH 16
+
+struct dns_entry {
+    char hostname[MAX_DNS_NAME];
+    char ip[MAX_IP_LENGTH];
+
```
scareything, line 30: Should there be a flag here to indicate the record type (A vs AAAA)?

lib/ziti_tunneler.c
```diff
@@ -19,6 +19,10 @@ limitations under the License.
 #define LWIP_DONT_PROVIDE_BYTEORDER_FUNCTIONS 1
 #endif
 
+#if defined(__mips) || defined(__mips__)
```
scareything, line 22: I'm curious why this is being ifdef'ed, since it was being defined on all architectures before this branch.

programs/ziti-tunnel-app/netif_driver/linux/tun.c
```diff
@@ -100,5 +100,8 @@ netif_driver tun_open(char *error, size_t error_len) {
     driver->uv_poll_init = tun_uv_poll_init;
     driver->close        = tun_close;
 
+    system("ip link set tun0 up");
+    system("ip route add 169.254.0.0/16 dev tun0");
```
scareything, line 104: You can use tun->name here instead of assuming "tun0".

scareything: I'm guessing you intend to use `dns_block` for the route destination?

But the more I think about this, the more uncertain I am about adding the route here. What happens when the tunneler app doesn't have/use a dns manager?

programs/ziti-tunnel-app/ziti-tunnel-app.c
```diff
@@ -32,28 +32,30 @@ void on_service(ziti_context ziti_ctx, ziti_service *service, int status, void *
...
 }
 
-static int run_tunnel() {
+extern dns_manager *get_dnsmasq_manager(const char* path);
+
+static int run_tunnel(const char *ip_range, dns_manager *dns) {
     uv_loop_t *ziti_loop = uv_default_loop();
     if (ziti_loop == NULL) {
-        fprintf(stderr, "failed to initialize default uv loop\n");
+        ZITI_LOG(ERROR, "failed to initialize default uv loop\n");
```
scareything, line 58: unneeded '\n' here and in a few other places.

lib/dns.c
ekoby: right now MAX_IP_LENGTH is 16, no support for IPv6 dns assignments. we can add those later

programs/ziti-tunnel-app/netif_driver/linux/tun.c
ekoby: I think we have to use DNS manager all the time (most of the time?) otherwise we can't configure DNS intercepts, right?

include/ziti/ziti_tunneler.h
ekoby: it is not used anywhere, the setting will be passed to global DNS mapper instead

lib/ziti_tunneler.c
ekoby: we never build for mips before, I think it could be because of the older gcc that have to be used

programs/ziti-tunnel-app/netif_driver/linux/tun.c
scareything: Most of the time, yes. But there are situations where the `--resolver none` option of the current ziti-tunnel for linux has been useful. Also consider at some point we may merge prox-c into the ziti-tunnel-c.

programs/ziti-tunnel-app/netif_driver/linux/tun.c
```diff
@@ -100,8 +100,12 @@ netif_driver tun_open(char *error, size_t error_len, const char *dns_block) {
     driver->uv_poll_init = tun_uv_poll_init;
     driver->close        = tun_close;
 
-    system("ip link set tun0 up");
-    system("ip route add 169.254.0.0/16 dev tun0");
+    char cmd[1024];
+    sprintf(cmd, "ip link set %s up", tun->name);
```
scareything, line 104: snprintf would be better.

programs/ziti-tunnel-app/netif_driver/linux/tun.c
ekoby: added a condition to add route only if DNS block is set

lib/dns.c
```diff
@@ -0,0 +1,71 @@
...
+const char* assign_ip(const char *hostname) {
+    struct dns_entry *e;
+    LIST_FOREACH(e, &ip_cache.entries, _next) {
+        if (strncmp(hostname, e->hostname, MAX_DNS_NAME) == 0) {
+            return e->ip;
+        }
+    }
+
+    e = calloc(1, sizeof(struct dns_entry));
+    uint32_t addr = ip_cache.base | (ip_cache.counter++ & ip_cache.counter_bits);
```
scareything, line 63: I just noticed this - I think it would be nice if assign_ip returned its argument if the argument is a valid IP address.

PR #54, opened 2020-10-26
implement half-close

ekoby:
implements TCP half-close support:
* intercepted connections - DONE
* hosted connections - TODO

### DIFF
--- a/include/ziti/ziti_tunnel_cbs.h
+++ b/include/ziti/ziti_tunnel_cbs.h
@@ -20,9 +20,11 @@ struct hosted_io_ctx_s {
     struct hosted_service_ctx_s *service;
     ziti_connection client;
     union {
-        uv_tcp_t *tcp;
-        uv_udp_t *udp;
+        uv_tcp_t tcp;
+        uv_udp_t udp;
     } server;
+    bool ziti_eof;
+    bool tcp_eof;
 };
 
 /** called by tunneler SDK after a client connection is intercepted */
--- a/include/ziti/ziti_tunnel_cbs.h
+++ b/include/ziti/ziti_tunnel_cbs.h
@@ -12,6 +12,8 @@ extern "C" {
 typedef struct ziti_io_ctx_s {
     ziti_connection      ziti_conn;
     tunneler_io_context  tnlr_io_ctx;
+    bool ziti_eof;
+    bool tnlr_eof;
 } ziti_io_context;
 
 struct hosted_io_ctx_s {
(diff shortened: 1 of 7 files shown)


include/ziti/ziti_tunnel_cbs.h
```diff
@@ -12,6 +12,8 @@ extern "C" {
 typedef struct ziti_io_ctx_s {
     ziti_connection      ziti_conn;
     tunneler_io_context  tnlr_io_ctx;
+    bool ziti_eof;
```
scareything, line 15: It feels like ziti_eof and tnlr_eof should belong to ziti_io_context and tnlr_io_context respecitively.

lib/ziti_tunnel_cbs.c
```diff
@@ -41,20 +41,41 @@ ssize_t on_ziti_data(ziti_connection conn, uint8_t *data, ssize_t len) {
...
     }
     return len;
 }
 
 /** called by tunneler SDK after a client connection is closed */
-void ziti_sdk_c_close(void *ziti_io_ctx) {
-    ziti_io_context *_ziti_io_ctx = ziti_io_ctx;
-    if (_ziti_io_ctx->ziti_conn != NULL) {
-        ziti_close(&_ziti_io_ctx->ziti_conn);
+int ziti_sdk_c_close(void *io_ctx) {
```
scareything, line 62: I'd prefer to name this arg `ziti_io_ctx` or similar. io_ctx could mislead someone to pass a struct io_ctx_s *.

scareything (approved):

include/ziti/ziti_tunnel_cbs.h
ekoby: It is easier here because it is the thing that glues them together and has to decide what to do, also `tunneler_io_context` is opaque to `ziti_tunnel_cbs` code and would require accessor or exposed internals

lib/tunnel_tcp.c
```diff
@@ -164,12 +194,28 @@ void tunneler_tcp_ack(struct write_ctx_s *write_ctx) {
     pbuf_free(wr_ctx->pbuf);
 }
 
+int tunneler_tcp_close_write(struct tcp_pcb *pcb) {
+    if (pcb == NULL) {
+        ZITI_LOG(WARN, "null pcb");
+        return 0;
+    }
+    LOG_STATE(DEBUG, "closing write", pcb);
+    err_t err = tcp_shutdown(pcb, 0, 1);
```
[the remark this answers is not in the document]
ekoby: `tcp_shutdown` avoids double-close internally and returns an error if state is not expected. If we get that error the problem is with our connection tracking logic.

PR #56, opened 2020-10-28
update github workflow

dovholuknf:
noticed the windows build was crazy fast - it wasn't actually building

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -20,14 +20,23 @@ jobs:
           fetch-depth: 0
       - name: make build dir
         run: cmake -E make_directory ${{runner.workspace}}/build
+        
+        #mac/linux
       - name: configure cmake
+        if: matrix.os != 'windows-latest'
         run: cmake -S ${{ github.workspace }} -B ${{runner.workspace}}/build
       - name: build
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -21,13 +21,13 @@ jobs:
       - name: make build dir
         run: cmake -E make_directory ${{runner.workspace}}/build
       - name: configure cmake
-        if: matrix.os != 'windows-latest'
         run: cmake -S ${{ github.workspace }} -B ${{runner.workspace}}/build
-      - name: configure cmake (WIN)
-        if: matrix.os == 'windows-latest'
-        run: cmake -S ${{ github.workspace }} -B ${{runner.workspace}}/build -G "MinGW Makefiles"
```
ekoby, line 28: why not MinGW? CGO requires MinGW build IIRC

dovholuknf: i didn't see the failure locally so i am changing this to be the same as to how i ended up building it from the go script. that uses `-G Ninja` which uses mingw

Also I was just hoping to see it compile since the go project buidls this each time anyway

scareything (approved):

PR #60, opened 2020-11-06
Avoid "altered pbuf payload" assertion when TCP connection limit is exceeded

### LINKED ISSUE #59: tunneler SDK crashes when TCP connection limit (MEMP_NUM_TCP_PCB) is reached
After failing to allocate a tcp pcb for an intercepted connection, the following assertion is hit in lwip:

    Assertion "raw pcb recv callback altered pbuf payload pointer without eating packet"

Avoid the assertion by returning 1 from recv_tcp (since the packet is consumed even though we can't intercept it).

Also expose the LWIP constants to the tunneler application build so platform-specific limits can be set.

scareything:
Also expose LWIP constants to tunneler build.

closes #59


dovholuknf (approved):

programs/ziti-edge-tunnel/CMakeLists.txt
```diff
@@ -17,6 +17,12 @@ if(CMAKE_SYSTEM_NAME STREQUAL Darwin OR CMAKE_SYSTEM_NAME STREQUAL Linux)
             PRIVATE ZITI_LOG_PREFIX=${PROJECT_NAME}
             )
 
+    # override LWIP defaults here
+    target_compile_definitions(lwipcore
+            PUBLIC MEMP_NUM_TCP_PCB=256
```
ekoby, line 22: we should add CMAKE options for these
`$ cmake -DMAX_TCP_CONNENCTIONS=1024 ...`

PR #64, opened 2020-11-16
make 100.64.0.0/10 default DNS assignement range

ekoby:


lib/dns.c
```diff
@@ -27,21 +27,19 @@ limitations under the License.
...
     uint32_t base;
     uint32_t counter;
     uint32_t counter_bits;
-    LIST_HEAD(entries, dns_entry) entries;
+    model_map entries;
 } cache;
 
 static cache ip_cache = {
-        .base = 0xA9FE0000, // 169.254.0.0
+        .base = 0x64400000, // 100.64.0.0
```
dovholuknf, line 40: saw this in wireguard code recently and figured i'd steal it some day... 
```
(100 << 24) | (64 << 16) | (0 << 8) | (0 << 0)
``` just a thought :)

File: lib/dns.c [hunk 2]
```diff
@@ -27,21 +27,19 @@ limitations under the License.
...
         .counter = 0x00000201, // 0.0.1.1 -- starting
-        .counter_bits = 0xffff,
+        .counter_bits = 0x3fffff, // 10-bit netmask
```
dovholuknf, line 42: i think this is just too darn big myself.... doesn't really matter i suppose but still. ¯\_(ツ)_/¯

dovholuknf (approved):

lib/dns.c
ekoby: bahh, just learn reading hex

PR #76, opened 2020-12-22
update ziti-sdk-c

ekoby:


dovholuknf:
pls don't merge this till after the new year to be safe

smilindave26 (approved):

lib/ziti_tunnel_cbs.c
```diff
@@ -448,4 +455,8 @@ void ziti_sdk_c_host_v1(void *ztx, uv_loop_t *loop, const char *service_name, co
     ziti_connection serv;
     ziti_conn_init(ziti_ctx, &serv, service_ctx);
     ziti_listen(serv, service_name, hosted_listen_cb, on_hosted_client_connect);
+}
+
+static void ziti_conn_close_cb(ziti_connection zc) {
```
scareything, line 460: Can the data_cb be called after ziti_close is called?

If so we would need to move the free'ing / closing of any io context members to the close_cb.

scareything (approved):

lib/ziti_tunnel_cbs.c
ekoby: we probably should move some of the cleanup into close_cb, however I didn't want to do it without evaluating the rest of the client lifecycle

PR #80, opened 2021-01-13
add toolchain for ARM

ekoby:


scareything (approved):
could you add a comment somewhere (I think the toolchain file would be fine) that says which host OS you were building on and how you installed the cross compiler and related tools?

PR #83, opened 2021-01-20
close cleanup

scareything:
* clear ziti conn data when ziti connection is closed
* always free pbuf in tcp data callback


ekoby (approved):
LGTM

PR #86, opened 2021-01-22
Close hosted connection when server sends EOF

scareything:


dovholuknf (approved):

lib/ziti_tunnel_cbs.c
```diff
@@ -279,6 +284,8 @@ static void on_hosted_tcp_server_data(uv_stream_t *stream, ssize_t nread, const
                 uv_close((uv_handle_t *) &io_ctx->server.tcp, hosted_server_close_cb);
             } else {
                 ziti_close_write(io_ctx->client);
+                uv_read_stop((uv_stream_t *) &io_ctx->server.tcp);
```
ekoby, line 287: pretty sure this is not necessary -- you won't get any more data after UV_EOF

ekoby (approved):

PR #112, opened 2021-02-25
multiple identity support

ekoby:
new `ziti-edge-tunnel` options
`-i|--identity` can be repeated multiple times to load multiple identities
`-I|--identity-dir` can be used to load all identities from a directory

### DIFF
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -15,15 +15,27 @@
 #error "please port this file to your operating system"
 #endif
 
+extern dns_manager *get_dnsmasq_manager(const char* path);
+
+struct ziti_instance_s {
+    ziti_options opts;
+    ziti_context ztx;
+    LIST_ENTRY(ziti_instance_s) _next;
+};
+
+// temporary list to pass info between parse and run
+static LIST_HEAD(instance_list, ziti_instance_s) instance_init_list;
+
+// map<path -> ziti_instance>
+static model_map instances;
+
 static void on_ziti_event(ziti_context ztx, const ziti_event_t *event);
 
 const char *cfg_types[] = { "ziti-tunneler-client.v1", "intercept.v1", "ziti-tunneler-server.v1", "host.v1", NULL };
-static ziti_options OPTS = {
-        .config_types = cfg_types,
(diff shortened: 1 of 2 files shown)


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -15,15 +15,27 @@
...
 
 const char *cfg_types[] = { "ziti-tunneler-client.v1", "intercept.v1", "ziti-tunneler-server.v1", "host.v1", NULL };
-static ziti_options OPTS = {
-        .config_types = cfg_types,
-        .events = ZitiContextEvent|ZitiServiceEvent,
-        .event_cb = on_ziti_event,
-        .refresh_interval = 10, /* default refresh */
-};
+
+static long refresh_interval = 10;
```
dovholuknf, line 36: 10 on purpose? shouldn't this be more like 300?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -164,14 +197,21 @@ static int run_opts(int argc, char *argv[]) {
     while ((c = getopt_long(argc, argv, "i:v:r:d:n:",
                             run_options, &option_index)) != -1) {
         switch (c) {
-            case 'i':
-                OPTS.config = strdup(optarg);
+            case 'i': {
```
smilindave26, line 200: what do you think about supporting a directory (and attempt to load all of the .json identity files in the dir).
Or a single file that contains a list of identity files (and other config info).
possible to add an identity without restarting? eventual IPC support, I guess...

dovholuknf: possibly even having it be `~/.ziti/identities` might be nice. the good thing is that it makes it easy to move your identity 
 for test reasons... which is the bad thing too possibly.  ZDEW works like this and scans the folder for 'new' identities when it starts up

ekoby: directory option is  coming

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
ekoby: this was the default, I am not changing that

dovholuknf (approved):

PR #127, opened 2021-03-19
Tunneler commands

ekoby:


include/ziti/ziti_tunnel_cbs.h
```diff
@@ -13,6 +13,40 @@ XX(data, string, map, data, __VA_ARGS__)
...
+
+#define TUNNEL_CMD_RES(XX, ...) \
+XX(success, bool, none, success, __VA_ARGS__) \
+XX(error, string, none, error, __VA_ARGS__)\
+XX(data, json, none, data, __VA_ARGS__)
+
+#define LOAD_IDENTITY(XX, ...) \
+XX(path, string, none, path, __VA_ARGS__)
+
+#define IDENTITY_INFO(XX, ...) \
```
scareything, line 35: Are you opposed to naming these so that it's clear they are tunnel commands? e.g. `TUNNEL_CMD_IDENTITY_INFO`, `TUNNEL_CMD_IDENTITY_LIST`?

lib/ziti_tunnel_ctrl.c
```diff
@@ -0,0 +1,245 @@
...
+    switch (cmd->command) {
+        case TunnelCommand_LoadIdentity: {
+            load_identity_cmd load;
+            if (cmd->data == NULL || parse_load_identity_cmd(&load, cmd->data, strlen(cmd->data)) != 0) {
+                result.success = false;
+                result.error = "invalid command";
+                break;
+            }
+            load_identity(load.path, cb, ctx);
+            return 0;
```
scareything, line 80: should `cb` be called here?

ekoby: It is called async after identity loads

lib/ziti_tunnel_ctrl.c
```diff
@@ -0,0 +1,245 @@
...
+}
+
+static int process_cmd(const tunnel_comand *cmd, command_cb cb, void *ctx) {
+    tunnel_result result = {
+                    .success = false,
+                    .error = NULL,
+                    .data = NULL,
+    };
+    ZITI_LOG(INFO, "processing command[%s] with data[%s]", TunnelCommands.name(cmd->command), cmd->data);
+    switch (cmd->command) {
```
scareything, line 71: should this switch have a case for `TunnelCommand_ZitiDump`?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -17,107 +18,101 @@
...
     }
-    free(config_path);
-    uv_close((uv_handle_t *) ar, (uv_close_cb) free);
+    static char sockfile[] = "/tmp/ziti-edge-tunnel.sock";
+    uv_fs_t fs;
+    uv_fs_unlink(l, &fs, sockfile, NULL);
+
+    uv_pipe_init(l, &cmd_server, 0);
+    uv_pipe_bind(&cmd_server, sockfile);
+    uv_pipe_chmod(&cmd_server, UV_WRITABLE | UV_READABLE);
```
scareything, line 112: Can we chown the socket file to prevent unprivileged access?

As it is now, any user on the box can load any identity they have access to, possibly from their own controller, and influence dns lookups.

lib/ziti_tunnel_ctrl.c
```diff
@@ -0,0 +1,245 @@
...
+                const ziti_identity *identity = ziti_get_identity(inst->ztx);
+                info->name = strdup(identity->name);
+                info->id = strdup(identity->id);
+                info->network = strdup(ziti_get_controller(inst->ztx));
+                info->config = strdup(key);
+
+                id_list.identities[idx++] = info;
+            }
+
+            result.data = identity_list_to_json(&id_list, MODEL_JSON_COMPACT, NULL);
```
scareything, line 100: is this the right place to set `result.success` to true?

currently seeing this:
```
$ echo -n '{ "command":"ListIdentities" }' | ncat -U /tmp/ziti-edge-tunnel.sock 
{"success":false,"data":{"identities":[{"name":"scarey@fedora-33-vm","config":"/media/psf/Home/.ziti/scarey@fedora-33-vm.json","network":"https://lambda-caeli.localdomain:11281","id":"jUSKpnLSx"},{"name":"mattermost.shawn.carey.pc","config":"/media/psf/Home/.ziti/mattermost.shawn.carey.pc_Enrollment.json","network":"https://3.210.34.143:443","id":"9pv6IQQYC"}]}}
```

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -17,107 +18,101 @@
...
-            model_map_set(&instances, config_path, inst);
-        } else {
-            ZITI_LOG(ERROR, "failed to initialize ziti");
-        }
+    if (uv_is_active((const uv_handle_t *) &cmd_server)) {
+        return 0;
     }
-    free(config_path);
-    uv_close((uv_handle_t *) ar, (uv_close_cb) free);
+    static char sockfile[] = "/tmp/ziti-edge-tunnel.sock";
```
scareything, line 106: I know it isn't something that we want to do very often, but it is currently possible to run multiple ziti-edge-tunnels on the same host. This might be more useful if/when we support host mode... Anyway, what do you think about putting the PID in the socket file name?

lib/ziti_tunnel_ctrl.c
ekoby: sure, once we implement it

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
ekoby: instead of pid tun name would be better I think -- `/tmp/ziti-edge-tunnel.tun0.sock`
we'll do it later when we know how UI can deal with it

scareything: I thought about tun name, but there won't be a tun device in some cases (e.g. if we support tproxy or something like host-only mode).

scareything (approved):

PR #128, opened 2021-03-19
free pbuf for dns query before query goes onto work queue

scareything:
Don't retain the pbuf for dns queries for the entire duration of the query.

### DIFF
--- a/lib/ziti_tunnel.c
+++ b/lib/ziti_tunnel.c
@@ -45,7 +45,6 @@ const char *CLIENT_PORT_KEY = "client_port";
 const char *SOURCE_IP_KEY = "source_ip";
 
 struct resolve_req {
-    struct pbuf *qp;
     ip_addr_t addr;
     u16_t port;
     tunneler_context tnlr_ctx;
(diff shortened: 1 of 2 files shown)


dovholuknf (approved):

smilindave26:
Update to latest C SDK while you're in there?

lib/ziti_tunnel.c
```diff
@@ -174,17 +172,16 @@ static void on_dns_packet(void *arg, struct udp_pcb *pcb, struct pbuf *p,
     tunneler_context tnlr_ctx = arg;
 
     struct resolve_req *rr = calloc(1,sizeof(struct resolve_req));
-    rr->qp = p;
     rr->addr = *addr;
     rr->port = port;
     rr->tnlr_ctx = tnlr_ctx;
 
     int rc = tnlr_ctx->dns->query(tnlr_ctx->dns, p->payload, p->len, send_dns_resp, rr);
```
ekoby, line 179: this is not safe, there is no guarantee that query is a synchronous operation, `fallback_cb` is always called on a worker thread.

You need to copy payload into `resolve_req`

ekoby (changes requested):
need to copy payload into resolve_req

ekoby (approved):

PR #138, opened 2021-04-05
add Docker build assets for ziti-edge-tunnel

qrkourier:
Still working on docs

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,6 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
-    needs: tag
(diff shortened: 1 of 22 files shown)


dovholuknf:
> can you add a readme on how to use?

lol - i see this above now: "Still working on docs", i'll await these :)

scareything (approved):
I would guess that eventually we want to bring the changes from https://github.com/openziti/ziti/pull/370 into the ziti-edge-tunnel docker image?

We might consider a base image, or at least a common repo, that contains startup/build scripts that are (or could be) shared between the ziti-edge-tunnel and ziti-tunnel images.

PR #141, opened 2021-04-19
Ziti dump on signal

ekoby:


lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -232,6 +234,9 @@ static void load_ziti_async(uv_async_t *ar) {
 }
 
 static void on_sigdump(uv_signal_t *sig, int signum) {
+#ifndef MAXPATHLEN
+#define MAXPATHLEN 1024
+#endif
     char fname[MAXPATHLEN];
     snprintf(fname, sizeof(fname), "/tmp/ziti-dump.%d.dump", getpid());
```
scareything, line 241: I'm not sure what this will do on Windows, but I'm guessing best-case is you end up with c:\tmp. Should we abstract the tmp directory so it's correct on each platform?

ekoby: there is no signals on Windows anyway, some are simulated by libuv(SIGTERM, SIGABRT) so this code is moot on win32 for now

scareything (approved):

PR #150, opened 2021-04-28
resolve $tunneler_id.name in listenOpts (0.15.x)

scareything:


lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -825,6 +824,15 @@ host_ctx_t *ziti_sdk_c_host(void *ziti_ctx, uv_loop_t *loop, const char *service
     ziti_connection serv;
     ziti_conn_init(ziti_ctx, &serv, host_ctx);
 
+    char listen_identity[128];
+    if (listen_opts_p != NULL) {
+        if (listen_opts_p->identity != NULL && listen_opts_p->identity[0] != '\0') {
+            const ziti_identity *zid = ziti_get_identity(ziti_ctx);
+            strncpy(listen_identity, listen_opts_p->identity, sizeof(listen_identity));
+            string_replace(listen_identity, sizeof(listen_identity), "$tunneler_id.name", zid->name);
+            listen_opts_p->identity = listen_identity;
```
dovholuknf, line 833: should this check the replaced value for NULL before assigning it? in the rare case that the memory overflows string_replace returns a NULL... should that be assigned here?

dovholuknf: actually looking at the function it returns a char* but `string_replace` doesn't seem to be using the returned value and i didn't see quite how it was manipulating the passed in string (`listen_identity`)...  maybe i'm just 'missing it'?

scareything: string_replace operates on the source string and returns a pointer to the first replaced character or null if no replacement was made.

I’ll change the logic to only replace the identity pointer if the return value of string_replace is not null.

PR #151, opened 2021-04-30
Win tun fix

ekoby:


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -110,16 +110,35 @@ static int start_cmd_socket(uv_loop_t *l) {
     if (uv_is_active((const uv_handle_t *) &cmd_server)) {
         return 0;
     }
+#if _WIN32
+    static char sockfile[] = "\\\\.\\pipe\\ziti-edge-tunnel.sock";
```
dovholuknf, line 114: it might be worthwhile to namespace these just for clarity's sake. I'd also used a pipe for one-way events (from the service to a consumer). just a passing thought/consideration

dovholuknf (approved):

PR #155, opened 2021-05-10
Ziti dump function

mary-dcouto:


lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -57,6 +57,70 @@ const ziti_tunnel_ctrl* ziti_tunnel_init_cmd(uv_loop_t *loop, tunneler_context t
...
+
+int ziti_dump_to_log_cb(void* stringsBuilder, const char *fmt,  ...) {
+    static char line[4096];
+
+    va_list vargs;
+    va_start(vargs, fmt);
+    vsnprintf(line, sizeof(line), fmt, vargs);
+    va_end(vargs);
+
+    if (strlen(stringsBuilder) + strlen(line) > 8192) {
```
scareything, line 69: 8192 should be a macro or static const so it can be reused in multiple functions.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -57,6 +57,70 @@ const ziti_tunnel_ctrl* ziti_tunnel_init_cmd(uv_loop_t *loop, tunneler_context t
...
+}
+
+void ziti_dump_to_log(void *ctx) {
+    char* buffer;
+    buffer = malloc(8192*sizeof(char));
+    buffer[0] = 0;
+    //actually invoke ziti_dump here
+    ziti_dump(ctx, ziti_dump_to_log_cb, buffer);
+    ZITI_LOG(INFO, "ziti dump to log %s", buffer);
+    free(buffer);
```
scareything, line 84: Is it safe to free the buffer here? I thought it would be used (and freed) in the callbacks, which happen asynchronously.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -57,6 +57,70 @@ const ziti_tunnel_ctrl* ziti_tunnel_init_cmd(uv_loop_t *loop, tunneler_context t
...
+    char time_str[32];
+    struct tm* start_tm = gmtime(&dump_time.tv_sec);
+    strftime(time_str, sizeof(time_str), "%FT%T", start_tm);
+
+    fprintf(fp, "Ziti Dump starting: %s\n",time_str);
+
+    //actually invoke ziti_dump here
+    ziti_dump(ctx, ziti_dump_to_file_cb, fp);
+    fflush(fp);
+    fclose(fp);
```
scareything, line 121: Similar to freeing the buffer in `ziti_dump_to_log`, here I think the file needs to remain open until `ziti_dump_to_file_cb` is invoked and completed. The invocation is asynchronous to this function.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
scareything: yeah, I looked at `ziti_dump` and it isn't creating any uv handles. Naming the function with `_cb` here is misleading though.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -492,6 +496,116 @@ static void enroll(int argc, char *argv[]) {
...
+    static struct option opts[] = {
+            {"identity", optional_argument, NULL, 'i'},
+            {"dump_path", optional_argument, NULL, 'p'},
+    };
+    int c, option_index, errors = 0;
+    optind = 0;
+
+    tunnel_ziti_dump *dump_options;
+    dump_options = calloc(1, sizeof(dump_options));
+    cmd = calloc(1, sizeof(tunnel_comand));
```
scareything, line 511: when does cmd get freed?

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -57,6 +57,70 @@ const ziti_tunnel_ctrl* ziti_tunnel_init_cmd(uv_loop_t *loop, tunneler_context t
...
+    ZITI_LOG(INFO, "ziti dump to log %s", buffer);
+    free(buffer);
+}
+
+int ziti_dump_to_file_cb(void* fp, const char *fmt,  ...) {
+    static char line[4096];
+
+    va_list vargs;
+    va_start(vargs, fmt);
+    vsnprintf(line, sizeof(line), fmt, vargs);
```
scareything, line 92: I think you could omit the fputs (and the temp buffer) and write to fp directly via vfprintf:

    vfprintf(fp, fmt, vargs);

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -492,6 +496,116 @@ static void enroll(int argc, char *argv[]) {
...
+
+    while ((c = getopt_long(argc, argv, "i:p:",
+                            opts, &option_index)) != -1) {
+        switch (c) {
+            case 'i':
+                dump_options->id = malloc(sizeof(*optarg));
+                dump_options->id = optarg;
+                break;
+            case 'p':
+                dump_options->dump_path = malloc(sizeof(*optarg));
```
ekoby, line 522: this malloc is not needed and is a memory leak

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -492,6 +496,116 @@ static void enroll(int argc, char *argv[]) {
...
+        send_message_to_pipe(connect, json, json_len);
+        free(json);
+
+    }
+}
+
+static uv_loop_t* connect_and_send_cmd(char sockfile[],uv_connect_t* connect, uv_pipe_t* client_handle) {
+    uv_loop_t* loop = uv_default_loop();
+
+    assert(0 == uv_pipe_init(loop, client_handle, 0));
```
ekoby, line 592: `assert` does nothing in release builds. Just print the error and exit instead

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -492,6 +496,116 @@ static void enroll(int argc, char *argv[]) {
...
+    tunnel_ziti_dump *dump_options;
+    dump_options = calloc(1, sizeof(dump_options));
+    cmd = calloc(1, sizeof(tunnel_comand));
+    cmd->command = TunnelCommand_ZitiDump;
+
+    while ((c = getopt_long(argc, argv, "i:p:",
+                            opts, &option_index)) != -1) {
+        switch (c) {
+            case 'i':
+                dump_options->id = malloc(sizeof(*optarg));
```
ekoby, line 518: same as below

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -492,6 +496,116 @@ static void enroll(int argc, char *argv[]) {
...
+    assert(0 == uv_pipe_init(loop, client_handle, 0));
+
+    uv_pipe_connect(connect, client_handle, sockfile, on_connect);
+
+    return loop;
+}
+
+static void dump(int argc, char *argv[]) {
+
+    uv_pipe_t* client_handle = (uv_pipe_t*)malloc(sizeof(uv_pipe_t));
```
ekoby, line 601: these could be on stack here

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
mary-dcouto: freed cmd in the on_connect function

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
scareything: I misunderstood the way ziti_dump is using the printer function. It's entirely synchronous, so the file and memory resources can be cleaned up here.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
mary-dcouto: renamed the function from ziti_dump_to_log_cb to ziti_dump_to_log_op

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
scareything: Ok, I see now that the `cmd` pointer is a static variable in this file and that it gets freed in on_connect. Now I question whether it should be freed? What happens to `cmd` when subsequent connections are made?

andrewpmartinez (approved):

andrewpmartinez (commented):
No text

andrewpmartinez (changes requested):
I hit approve when I didn't mean to. Apparently there is no way to remove approval?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -495,6 +498,128 @@ static void enroll(int argc, char *argv[]) {
...
+        exit(1);
+    }
+    size_t json_len;
+    cmd->data = tunnel_ziti_dump_to_json(dump_options, MODEL_JSON_COMPACT, &json_len);
+
+    return optind;
+}
+
+static void close_cmd_socket(uv_pipe_t* handle) {
+    uv_unref(handle);
```
ekoby, line 542: no need to `unref` before closing

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -495,6 +498,128 @@ static void enroll(int argc, char *argv[]) {
...
+    uv_close(handle, NULL);
+}
+
+static void on_response(uv_stream_t *s, ssize_t len, const uv_buf_t *b) {
+    if (len > 0) {
+        printf("received response <%.*s>\n", (int) len, b->base);
+    } else {
+        printf("Read Response error %s\n", uv_err_name(len));
```
ekoby, line 550: write errors to `stderr`

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -495,6 +498,128 @@ static void enroll(int argc, char *argv[]) {
...
+        printf("received response <%.*s>\n", (int) len, b->base);
```
ekoby, line 551: can't do this if `len < 0` and buffer would contain garbage,

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -495,6 +498,128 @@ static void enroll(int argc, char *argv[]) {
...
+    if (status < 0) {
+        printf("Could not sent message to the tunnel. Write error %s\n", uv_err_name(status));
+    } else {
+        puts("Message sent to the tunnel.");
+    }
+    free(req);
+}
+
+void send_message_to_pipe(uv_connect_t *connect, void *message, size_t datalen) {
+    printf("Message...%s\n", message);
```
ekoby, line 567: is this for debugging? should use ZITI_LOG then

mary-dcouto: This message will be displayed in the command prompt, when you execute the ziti-edge-tunnel dump command from the commandline. So I used printf statements. I thought ZITI_LOG is used for writing to the log file.

ekoby (approved):

PR #165, opened 2021-06-16
Throttle hosting connections

ekoby:


lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -205,14 +213,45 @@ static ssize_t on_hosted_client_data(ziti_connection clt, uint8_t *data, ssize_t
...
+    if (len < 0) {
+        ZITI_LOG(WARN, "ziti write error: %zd(%s), stop reading peer", len, ziti_errorstr((int)len));
+
+        switch (io_ctx->server_proto_id) {
+            case IPPROTO_TCP:
+                uv_read_stop((uv_stream_t *) &io_ctx->server.tcp);
+                break;
+            case IPPROTO_UDP:
+                uv_udp_recv_stop(&io_ctx->server.udp);
+                break;
```
dovholuknf, line 250: does it makes sense to have a 'guard' case should we ever do something that is not tcp/udp?

lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -34,6 +34,8 @@ limitations under the License.
 #endif
 #endif
 
+#define ZITI_MTU (15 * 1024)
```
scareything, line 37: 15k? Does this relate to a ziti or uv parameter?

ekoby: that was Ziti limit

dovholuknf: i was going to ask if this was the source of the issue olaf reported in discourse. it was just moved from down below though

lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -205,14 +213,45 @@ static ssize_t on_hosted_client_data(ziti_connection clt, uint8_t *data, ssize_t
...
 
-#define ZITI_MTU (15 * 1024)
 static void alloc_buffer(uv_handle_t *handle, size_t suggested_size, uv_buf_t *buf) {
-    *buf = uv_buf_init((char*) malloc(ZITI_MTU), ZITI_MTU);
+    struct hosted_io_ctx_s *io_ctx = handle->data;
+
+    if (io_ctx->in_wreqs < MAX_OUTSTANDING_WRITES) {
+        *buf = uv_buf_init((char*) malloc(ZITI_MTU), ZITI_MTU);
+    } else {
+        ZITI_LOG(VERBOSE, "max ziti writes[%d] reached for %s", io_ctx->in_wreqs, io_ctx->server_dial_str);
```
dovholuknf, line 222: this could destroy the disk if verbose logging is enabled and a mountain of data is trying to be sent. seems DOS-like to me. can it be throttled?

scareything (approved):

lib/ziti-tunnel-cbs/ziti_hosting.c
ekoby: this is on connection that ziti-edge-tunnel initiates, also this is no different from any other place we do verbose logging

PR #171, opened 2021-07-09
Fix hosted connection close

ekoby:
Ziti Connection was not torn down if there were outstanding ziti write requests at tcp.eof


dovholuknf (approved):
seems good to me! :D

PR #172, opened 2021-07-12
exclude ziti routers from tunnel intercept (force route via default gw)

ekoby:


programs/ziti-edge-tunnel/netif_driver/linux/tun.c
```diff
@@ -181,6 +181,22 @@ static void init_dns_maintainer(uv_loop_t *loop, const char *tun_name, uint32_t
     do_dns_update(loop, 0);
 }
 
+static int tun_exclude_rt(netif_handle dev, uv_loop_t *l, const char *addr) {
```
scareything, line 184: do we need a function (or flag) to un-exclude routes?

programs/ziti-edge-tunnel/netif_driver/linux/tun.c
```diff
@@ -181,6 +181,22 @@ static void init_dns_maintainer(uv_loop_t *loop, const char *tun_name, uint32_t
...
+    ZITI_LOG(DEBUG, "default route is '%.*s'", def_rt_size, def_route);
+    pclose(def_rt);
+
+    char *p = strstr(def_route, "via ");
+    if (p == NULL) return -1;
+    char *gw = p + 4;
+    char *endgw = strchr(gw, ' ');
+    int gw_len = (int)(endgw - gw);
+
+    return run_command("ip route replace %s via %.*s", addr, gw_len, gw);
```
scareything, line 197: I think we also need to track when the default gateway changes.

scareything (approved):

lib/ziti-tunnel/include/ziti/netif_driver.h
```diff
@@ -17,6 +17,7 @@ typedef int (*uv_poll_req_fn)(netif_handle dev, uv_loop_t *loop, uv_poll_t *tun_
 typedef int (*setup_packet_cb)(netif_handle dev, uv_loop_t *loop, packet_cb cb, void *netif);
 typedef int (*add_route_cb)(netif_handle dev, const char *dest);
 typedef int (*delete_route_cb)(netif_handle dev, const char *dest);
+typedef int (*exclude_route_fn)(netif_handle dev, uv_loop_t *loop, const char *dest);
```
scareything, line 20: What do you think about renaming this to `add_address_cb`? I think it better reflects what the function actually does (rather than why it is being done), and we need to call the same function for hosted connections that set source IP.

scareything: never mind - I see route_exclusion is different from adding an address. I'll add another cb for add_address.

PR #169, opened 2021-06-22
Enable mfa cmd

mary-dcouto:

PR #177, opened 2021-07-16
Update ziti sdk c

ekoby:
MFA callback are gone, will be put back in #169

### DIFF
--- a/lib/ziti-tunnel-cbs/ziti_instance.h
+++ b/lib/ziti-tunnel-cbs/ziti_instance.h
@@ -29,7 +29,7 @@ static struct cmd_ctx_s {
 struct mfa_request_s {
     ziti_context ztx;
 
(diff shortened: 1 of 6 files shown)


lib/ziti-tunnel-cbs/ziti_instance.h
```diff
@@ -29,7 +29,7 @@ static struct cmd_ctx_s {
 struct mfa_request_s {
     ziti_context ztx;
 
-    ziti_ar_mfa_cb submit_f;
+    // ziti_ar_mfa_cb submit_f;
```
dovholuknf, line 32: actually delete it?

dovholuknf: same for the other commented out code...  maybe? or not ... :)

dovholuknf (approved):

PR #181, opened 2021-07-29
avoid using stale service_name reference

ekoby:


lib/ziti-tunnel/tunnel_tcp.c
```diff
@@ -258,7 +258,7 @@ static tunneler_io_context new_tunneler_io_context(tunneler_context tnlr_ctx, co
         return NULL;
     }
     ctx->tnlr_ctx = tnlr_ctx;
-    ctx->service_name = service_name;
```
scareything, line 261: We do this also in recv_udp().

ekoby: changed there as well

scareything (approved):

mary-dcouto (approved):

PR #182, opened 2021-08-09
Verify mfa cmd

mary-dcouto:


lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -301,7 +370,7 @@ static int load_identity(const char *identifier, const char *path, command_cb cb
 
 static struct ziti_instance_s *new_ziti_instance(const char *identifier, const char *path) {
     struct ziti_instance_s *inst = calloc(1, sizeof(struct ziti_instance_s));
-    inst->identifier = strdup(identifier ? identifier : path);
+    inst->identifier = strdup(identifier ? identifier : extract_filename(path));
```
ekoby, line 373: is there a good reason to drop the full path? makes it impossible to ids with same filename from different folders, for example:
* $HOME/.ziti/staging/id.json
* $HOME/.ziti/production/id.json

mary-dcouto: ok, I will remove the extract_filename function

ekoby (approved):

PR #185, opened 2021-08-11
Enhance docker entrypoint

qrkourier:


docker/docker-entrypoint.sh
```diff
@@ -1,56 +1,90 @@
...
+if ! mountpoint "${IDENTITIES_DIR}" &>/dev/null; then
+    echo "ERROR: please run this image with a volume mounted on ${IDENTITIES_DIR}" >&2
     exit 1
 fi
 
-# try to figure out the client name if it wasn't provided
-if [ -z "${NF_REG_NAME}" ]; then
-    if [ -n "${IOTEDGE_DEVICEID}" ]; then
-        echo "setting NF_REG_NAME to \${IOTEDGE_DEVICEID} (${IOTEDGE_DEVICEID})"
+# TODO: What is the origin story of IOTEDGE_DEVICEID and can it be decommissioned? -Ken
```
scareything, line 25: Azure IoT Edge modules have this variable set automatically, so this bit of code removes one potential failure when running as an IoT edge module. I don't think we've actually done that in a while though, so I'm ambivalent about removing it.

docker/docker-entrypoint.sh
```diff
@@ -1,56 +1,90 @@
...
+if [[ -z "${NF_REG_NAME:-}" ]]; then
```
scareything, line 26: How is this different than the vanilla Bourne shell conditional that uses `[`?

docker/docker-entrypoint.sh
```diff
@@ -1,56 +1,90 @@
...
+    if [[ -n "${IOTEDGE_DEVICEID:-}" ]]; then
+        echo "INFO: setting NF_REG_NAME to \${IOTEDGE_DEVICEID} (${IOTEDGE_DEVICEID})"
         NF_REG_NAME="${IOTEDGE_DEVICEID}"
     fi
 fi
-if [ -z "${NF_REG_NAME}" ]; then
```
scareything, line 28: I think this can only be removed if the IOTEDGE_DEVICEID logic is removed. This prevents falling through without an identity name.

scareything (approved):

docker/docker-entrypoint.sh
qrkourier: I swapped those out for the sake of uniformity of dialect, for lack of a better word. When I'm optimizing for portability I'll use a Bourne shell dialect that avoids modern Bashisms, but when I'm optimizing for safety and maintainability where I'm sure modern BASH is available then I'll prefer that. Let me know if you have any contradictory thoughts or preferences.

docker/docker-entrypoint.sh
qrkourier: The net effect of the changes is currently:

```bash
# pseudo-bash
if NF_REG_NAME; then
    identity is NF_REG_NAME
elif IOTEDGE_DEVICEID; then
    identity is IOTEDGE_DEVICEID
else
    false
fi
```
Do you think that'll work with the way you remember the Azure env var working?

docker/docker-entrypoint.sh
scareything, line 25: Is this comment still needed? The basic idea is to automatically set the identity to match the device ID of the iod edge module.

qrkourier: I've replaced that with a description of the var, thanks.

PR #191, opened 2021-08-16
Enhance docker build

qrkourier:


qrkourier:
I branched from v0.15.8 to hotfix the missing Docker assets at that point in the history. There wasn't a GitHub artifact for 0.15.8 so I added a second Dockerfile(.copy) to facilitate building the container image with a locally-compiled executable, and also to work around a current problem with `debian:stable-slim`, substituting `:buster-slim` (also Debian 10, if not identical to `:stable-slim`).

docker/buildx.sh
```diff
@@ -21,59 +21,65 @@ EXAMPLES
...
+docker run --rm --platform linux/arm64/v8 arm64v8/alpine uname -a |grep -Eq 'aarch64 Linux'
+docker run --rm --platform linux/arm/v7 arm32v7/alpine uname -a|grep -Eq 'armv7l Linux'
 docker buildx create --use --name=ziti-builder 2>/dev/null || docker buildx use --default ziti-builder
 
-docker buildx build $DIRNAME \
-    --platform linux/amd64,linux/arm/v7,linux/arm64 \
-    --build-arg ZITI_VERSION=${ZITI_VERSION} \
-    --tag ${CONTAINER_REPO}:${ZITI_VERSION} \
-    --tag ${CONTAINER_REPO}:latest \
+eval docker buildx build "${DIRNAME}" \
```
scareything, line 79: Why use `eval` here? Does the buildx output include a shell command that this shell is executing?

docker/buildx.sh
```diff
@@ -21,59 +21,65 @@ EXAMPLES
...
+    --platform="linux/amd64,linux/arm/v7,linux/arm64" \
+    --build-arg=ZITI_VERSION="${ZITI_VERSION}" \
+    "${TAG_PARAMS}" \
```
scareything, line 82: Oh, maybe `eval` is eating the quotes that surround TAG_PARAMS so multiple tags can be passed successfully? Could we instead not use `eval` here and omit the quotes around TAG_PARAMS?

scareything: The dockerfile change introduces a lot of duplication. The diff is between the two files is very small. In fact the only differences are the arguments to `COPY`.

    $ diff ~/tmp/dockerfile.fetch ~/tmp/dockerfile.copy
    17,19d16
    < COPY fetch-github-releases.sh .
    < RUN bash ./fetch-github-releases.sh ziti-edge-tunnel
    < 
    36d32
    < COPY --from=fetch-ziti-artifacts /tmp/ziti-edge-tunnel /usr/local/bin
    38c34,35
    < RUN chmod +x /docker-entrypoint.sh
    ---
    > COPY ./ziti-edge-tunnel /usr/local/bin/
    > RUN chmod +x /docker-entrypoint.sh ./usr/local/bin/ziti-edge-tunnel

I'm concerned that changes will be made to one dockerfile while the other is neglected.

scareything (approved):

docker/buildx.sh
qrkourier: @scareything I think it would be OK to drop `Dockerfile.copy` based on my current estimation of the reduced likelihood that we'll need to continue building container images for older versions of `ziti-edge-tunnel` that pre-date the earliest published releases in GitHub.

qrkourier: @scareything Re: quotes around TAG_PARAMS: that's correct. I used `eval` here just to avoid having a bare env var that's could be interpolated with arbitrary code as the value. I've been trying to be more intentional about quoting vars ever since I started using a shell script linter. It always points these out.

qrkourier: @scareything Let me know if your can think of another way to introduce variations on a Dockerfile, but I've removed the almost-redundant `Dockerfile.copy` from this branch because I don't think we really need it.

qrkourier:
@scareything I just added a commit to add the Dockerfile I forgot to include on Tuesday when I responded by dropping the unnecessary `Dockerfile.copy`.

PR #193, opened 2021-08-19
UDP timeout

### LINKED ISSUE #188: Connections established on behalf of UDP clients are never closed
Ensure that connections associated with UDP clients are closed when it appears that the UDP client has gone away. This could be implemented by starting/resetting a per-client timer when UDP packets are received; the associated ziti connection is closed when the timer expires.

ekoby:
[fixes #188]

### DIFF
--- a/lib/ziti-tunnel/tunnel_udp.c
+++ b/lib/ziti-tunnel/tunnel_udp.c
@@ -3,7 +3,7 @@
 #include "tunnel_udp.h"
 #include "ziti_tunnel_priv.h"
 
-#define UDP_TIMEOUT 30
+#define UDP_TIMEOUT 30000
 
 // initiate orderly shutdown
 static void udp_timeout_cb(uv_timer_t *t) {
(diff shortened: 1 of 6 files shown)


lib/ziti-tunnel/tunnel_udp.c
```diff
@@ -3,6 +3,14 @@
 #include "tunnel_udp.h"
 #include "ziti_tunnel_priv.h"
 
+#define UDP_TIMEOUT 30
```
scareything, line 6: This will probably end up needing to be configurable.

lib/ziti-tunnel/tunnel_udp.c
```diff
@@ -3,6 +3,14 @@
...
+
+// initiate orderly shutdown
```
scareything, line 8: Will it be ok if this happens while the connection has pending writes (e.g. waiting for router reconnect)?

scareything (approved):

lib/ziti-tunnel/tunnel_udp.c
ekoby: Timer is started/reset when actual payload is written to either direction

PR #202, opened 2021-09-06
Broadcast status message

mary-dcouto:
broadcast status message from Ziti edge tunnel


programs/ziti-edge-tunnel/instance.c
```diff
@@ -230,9 +234,11 @@ static void setTunnelServiceAddress(tunnel_service *tnl_svc, ziti_service *servi
         protocols = cfg_v1.protocols;
 
         // set ports
-        tnl_port_range_arr = calloc(sizeof(cfg_v1.port_ranges), sizeof(tunnel_port_range *)); // todo sizeof(cfg_v1.port_ranges) does not yield the number of port ranges.
-        int port_idx;
-        for(port_idx=0; cfg_v1.port_ranges[port_idx]; port_idx++) {
+        for(idx = 0; cfg_v1.port_ranges[idx]; idx++) {
```
ekoby, line 237: empty loop can be removed?

ekoby (approved):

programs/ziti-edge-tunnel/instance.c
mary-dcouto: I added empty loop, because I wanted to find the number of port ranges in the config

PR #200, opened 2021-09-03
Broadcast identity, mfa and service message

mary-dcouto:

PR #208, opened 2021-09-17
fix crashes

scareything:
My tunneler was crashing after pulling https://github.com/openziti/ziti-tunnel-sdk-c/pull/200. This PR gets things back to functional for the services and configs that I'm currently testing with.

I feel like a discussion about the service event model would be worthwhile. Without knowing more about the goals/intent of the service event model, it seems like we could eliminate some code if the events included the intercept_ctx instead of parsing the service configuration again.


mary-dcouto (approved):
Thanks for fixing the crash. I didnt have all the service configs, so I could not test all the combinations

ekoby (approved):

ekoby:
@mary-dcouto it might worth catching no event client condition earlier to avoid constructing these objects if we not going to send them anyway

mary-dcouto:
> @mary-dcouto it might worth catching no event client condition earlier to avoid constructing these objects if we not going to send them anyway

OK

PR #214, opened 2021-09-30
Misc fixes

ekoby:


programs/ziti-edge-tunnel/instance.c
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Netfoundry, Inc.
+Copyright 2019 NetFoundry, Inc.
```
dovholuknf, line 2: it's not supposed to have the comma fwiw... we need to clean all these up everywhere still... :(

dovholuknf (approved):

programs/ziti-edge-tunnel/instance.c
ekoby: CLion now has copyright update action!

PR #216, opened 2021-10-06
Controller events

mary-dcouto:


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -591,23 +606,33 @@ static void on_event(const base_event *ev) {
...
                     svc_event.RemovedServices[idx++] = svc;
                 }
+                svc_event.RemovedServices[idx] = NULL;
             }
 
             idx = 0;
             if (svc_ev->added_services != NULL) {
-                svc_event.AddedServices = calloc(sizeof(svc_ev->added_services), sizeof(tunnel_service *));
+                int svc_array_length = 0;
+                for (zs = svc_ev->added_services; *zs != NULL; zs++) {
```
scareything, line 627: as above, could you iterate with the index?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -591,23 +606,33 @@ static void on_event(const base_event *ev) {
...
+                    svc_array_length++;
+                }
+                svc_event.AddedServices = calloc(svc_array_length + 1, sizeof(tunnel_service *));
                 for (zs = svc_ev->added_services; *zs != NULL; zs++) {
                     tunnel_service *svc = get_tunnel_service(id, *zs);
                     svc_event.AddedServices[idx++] = svc;
                 }
+                svc_event.AddedServices[idx] = NULL;
```
scareything, line 635: redundant

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -616,26 +641,38 @@ static void on_event(const base_event *ev) {
 
             send_events_message(&svc_event, (to_json_fn) services_event_to_json, true);
             if (svc_event.AddedServices != NULL) {
+#if !defined(_WIN32)
```
scareything, line 644: This pointer needs to be free'd on Windows, since it is allocated unconditionally on all platforms. If you are seeing a crash when free'ing this on Windows then something else is causing the crash and it should be addressed.

mary-dcouto: yes, this is crashing in windows.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -592,21 +607,29 @@ static void on_event(const base_event *ev) {
...
+                    svc_array_length++;
+                }
+                svc_event.RemovedServices = calloc(svc_array_length + 1, sizeof(struct tunnel_service_s));
+                for (int svc_idx = 0; svc_ev->removed_services[svc_idx]; svc_idx++) {
+                    tunnel_service *svc = find_tunnel_service(id, svc_ev->removed_services[svc_idx]->id);
                     if (svc == NULL) {
-                        svc = get_tunnel_service(id, *zs);
+                        svc = get_tunnel_service(id, svc_ev->removed_services[svc_idx]);
                     }
                     svc_event.RemovedServices[idx++] = svc;
```
scareything, line 620: should this use `idx` or `svc_idx`? It seems like `idx` is not needed, at least in this loop?

mary-dcouto: fixed the index

PR #227, opened 2021-10-27
Config file backup

mary-dcouto:
creates a config.json and config.json.backup files in the APPDATA folder of the system user in windows and /tmp folder in linux. The tunnel will save identity information as soon as the identity files are loaded. So currently it does not save the information that are received after the mfa auth. The file will have the data in the same format as what WDE currently supports


programs/ziti-edge-tunnel/instance.c
```diff
@@ -34,14 +36,36 @@ tunnel_identity *find_tunnel_identity(const char* identifier) {
...
         return id;
     } else {
         tunnel_identity *tnl_id = calloc(1, sizeof(struct tunnel_identity_s));
         tnl_id->Identifier = strdup(identifier);
+        if (filename != NULL) {
+            char* extension = strstr(filename, ".json");
+
+            if (extension != NULL) {
+                int length = (int) (extension - filename);
+                tnl_id->FingerPrint = calloc(length + 1, sizeof(char));
```
ekoby, line 55: if `FingerPrint` just filename without `.json`? I don't understand the need for it

ekoby: Also it is not set if filename does not end with `.json`. This seems flaky

mary-dcouto: Fingerprint is the identifier that WDE UI / go tunnel code uses currently. We can remove it later, once all the references in the UI is changed to the Identifier variable. That may take time. So I thought, the fastest way to get this integrated with WDE UI, is to add Fingerprint variable as well in to the event messages

programs/ziti-edge-tunnel/config-utils.c
```diff
@@ -0,0 +1,64 @@
...
+
+const char* app_data = "APPDATA";
+static char* identifier_path = NULL;
+
+char* get_system_config_path() {
+    char* config_path = malloc(FILENAME_MAX * sizeof(char));
+#if _WIN32
+    sprintf(config_path, "%s/NetFoundry", getenv(app_data));
+#else
+    sprintf(config_path, "%s", "/tmp");
```
ekoby, line 29: this is bad place to store config: `/tmp` is wiped on reboot

mary-dcouto: I have moved config-utils.c to Windows in CmakeLists, I will fix this else part anyway.

programs/ziti-edge-tunnel/instance.c
mary-dcouto: fixed the issue with the null fingerprint value, when the file name does not end with .json

programs/ziti-edge-tunnel/config-utils.c
mary-dcouto: fixed the config path for __linux__ env. When we use this param for other platforms, we will fix it for other platforms. Right now, it is being used for windows only

programs/ziti-edge-tunnel/include/model/events.h
```diff
@@ -85,8 +87,12 @@ XX(normal, __VA_ARGS__) \
 XX(connected, __VA_ARGS__) \
 XX(disconnected, __VA_ARGS__)
 
+#define INSTANCE_STATUS(XX, ...) \
+XX(ok, __VA_ARGS__)
```
ekoby, line 91: shouldn't it have more than one value?

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,165 @@
...
+    char* tunnel_status = tunnel_status_to_json(stat, 0, &json_len);
+    bool saved = false;
+
+    if (json_len > 0) {
+        char* config_path = get_system_config_path();
+
+        char* config_file_name = get_config_file_name(config_path);
+        char* bkp_config_file_name = get_backup_config_file_name(config_path);
+
+        int stat = uv_mutex_trylock(&mutex);
```
ekoby, line 109: what is the mutex for?

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,165 @@
...
+        if (stat == 0) {
+            //copy config to backup file
+            FILE* config = fopen(config_file_name, "r");
+            if (config == NULL) {
+                ZITI_LOG(ERROR, "Could not open config file %s", config_file_name);
+            } else {
+                FILE* backup_config = fopen(bkp_config_file_name, "w");
```
ekoby, line 116: why not just name the file?

programs/ziti-edge-tunnel/include/model/events.h
mary-dcouto: I am using this variable just to check whether the identity file is present in the config path. WDE maintains a config.json file which will have the fingerprint, name and other details of the identities. When we load that file, we need to use a variable that indicates which identity has the corresponding fingerprint file.

programs/ziti-edge-tunnel/instance-config.c
mary-dcouto: if two threads try to write to the config file at the same time, then only one of them needs to succeed. The current status of the tunnel is always taken from the instance.c file, it is not passed into this function as data. So we don't want to write the data twice, we will discard one request.

programs/ziti-edge-tunnel/instance-config.c
mary-dcouto: changed the code, used rename instead of file overwrite operation

programs/ziti-edge-tunnel/include/model/dtos.h
```diff
@@ -80,7 +81,7 @@ XX(Addresses, tunnel_address, array, Addresses, __VA_ARGS__) \
 XX(Ports, tunnel_port_range, array, Ports, __VA_ARGS__)  \
 XX(OwnsIntercept, bool, none, OwnsIntercept, __VA_ARGS__) \
 XX(PostureChecks, tunnel_posture_check, array, PostureChecks, __VA_ARGS__) \
-XX(IsAccessible, bool, none, IsAccessible, __VA_ARGS__) \
+XX(IsAccessible, bool, none, IsAccessable, __VA_ARGS__) \
```
ekoby, line 84: we should fix Win UI instead of keeping using typos

programs/ziti-edge-tunnel/instance.c
```diff
@@ -354,7 +382,9 @@ tunnel_identity_array get_tunnel_identities() {
 
     int idx = 0;
     MODEL_MAP_FOREACH(id, tnl_id, &tnl_identity_map) {
-        tnl_id_arr[idx++] = tnl_id;
+        if (tnl_id->Status != NULL && strcmp(tnl_id->Status, instance_status_name(instance_status_ok)) == 0) {
```
ekoby, line 385: change tnl_id.Status to instance_status enum, then you don't need to use strcmp()

programs/ziti-edge-tunnel/include/model/dtos.h
mary-dcouto: fixed

programs/ziti-edge-tunnel/instance.c
mary-dcouto: changed the status field

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,151 @@
...
+        uv_mutex_lock(&mutex);
+        //copy config to backup file
+        int rem = remove(bkp_config_file_name);
+        if (rem == 0) {
+            ZITI_LOG(DEBUG, "Deleted backup config file %s", bkp_config_file_name);
+        }
+        if (rename(config_file_name, bkp_config_file_name) == 0) {
+            ZITI_LOG(DEBUG, "Copied config file to backup config file %s", bkp_config_file_name);
+        } else {
+            ZITI_LOG(DEBUG, "Could not copy config file [%s] to backup config file, the config might not exists at the moment", config_file_name);
```
scareything, line 118: Should this be logged at ERROR?

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        // write tunnel status to the config file
+        FILE* config = fopen(config_file_name, "w");
+        if (config == NULL) {
+            ZITI_LOG(ERROR, "Could not open config file %s to store the tunnel status data", config_file_name);
+        } else {
+            char* tunnel_status_data = tunnel_status;
+            for (int i =0; i< json_len; i=i+MIN_BUFFER_LEN-1, tunnel_status_data=tunnel_status_data+MIN_BUFFER_LEN-1) {
```
scareything, line 127: I may be missing something but ( think this entire loop could be replaced with:

    fputs(tunnel_status, config);

programs/ziti-edge-tunnel/instance.c
```diff
@@ -354,7 +382,9 @@ tunnel_identity_array get_tunnel_identities() {
 
     int idx = 0;
     MODEL_MAP_FOREACH(id, tnl_id, &tnl_identity_map) {
-        tnl_id_arr[idx++] = tnl_id;
+        if (tnl_id->Status == instance_status_ok) {
```
scareything, line 385: Why not move this test outside of the foreach, to avoid unnecessarily iterating the map?

    if (tnl_id->Status == instance_status_ok) {
        MODEL_MAP_FOREACH(id, tnl_id, &tnl_identity_map) {
            tnl_id_arr[idx++] = tnl_id;
        }
    }

programs/ziti-edge-tunnel/CMakeLists.txt
```diff
@@ -25,9 +25,9 @@ if (MSVC)
     set(getopt libwinc)
 endif()
 
-set(ZITI_INSTANCE_COMMON include/model/events.h include/model/dtos.h instance.c include/instance.h log_utils.c utils.c include/file-rotator.h)
+set(ZITI_INSTANCE_COMMON include/model/events.h include/model/dtos.h instance.c include/instance.h config-utils.c include/config-utils.h instance-config.c include/instance-config.h)
 if (WIN32)
-    set(ZITI_INSTANCE_OS windows-service.c include/windows/windows-service.h)
+    set(ZITI_INSTANCE_OS windows-service.c include/windows/windows-service.h log_utils.c include/log-utils.h)
```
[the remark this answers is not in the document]
mary-dcouto: I asked Eugene if we want to save the logs to the tmp folder in linux, and he told me that the linux users are supposed to use the journalctl to read the logs. So I didnt add this file for other platforms. If you want me to make this logutils available for all platforms, I can do that

programs/ziti-edge-tunnel/instance-config.c
mary-dcouto: this file will contain multiline data, and as of now it will have more than 1000 lines of data. So I thought, we have to write the data using buffer. If that is not required, I will change it as fputs

programs/ziti-edge-tunnel/instance.c
mary-dcouto: tnl_id is the value of each entry in the map, that is why I am checking this condition from inside the loop.

programs/ziti-edge-tunnel/instance-config.c
mary-dcouto: updated the code

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,151 @@
...
+
+bool load_tunnel_status_from_file(uv_loop_t* ziti_loop) {
+    char* config_path = get_system_config_path();
+
+    uv_fs_t fs;
+    int check = uv_fs_mkdir(ziti_loop, &fs, config_path, 0755, NULL);
+    if (check == 0) {
+        ZITI_LOG(TRACE,"config path is created at %s", config_path);
+    } else {
+        ZITI_LOG(TRACE,"config path is found at %s", config_path);
```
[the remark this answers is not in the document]
mary-dcouto: updated the code

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,151 @@
...
+    }
+    return mutex_init;
+}
+
+bool load_config_from_file(char* config_file_name) {
+    bool loaded = false;
+
+    FILE* config_file = fopen(config_file_name, "r");
+    if (config_file != NULL) {
+        char *config_buffer = malloc(MAX_BUFFER_LEN * sizeof(char));
```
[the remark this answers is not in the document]
mary-dcouto: updated the code

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,151 @@
...
+        *config_buffer = NULL;
```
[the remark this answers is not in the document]
mary-dcouto: updated the code

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,151 @@
...
+
+        if (strlen(config_buffer) > 0) {
+            loaded = load_tunnel_status(config_buffer);
+            if (!loaded) {
+                ZITI_LOG(WARN, "Config file %s cannot be read, will be overwritten", config_file_name);
+            }
+        }
+        fclose(config_file);
+    } else {
+        ZITI_LOG(INFO, "The config file %s does not exist. This is normal if this is a new install or if the config file was removed manually", config_file_name);
```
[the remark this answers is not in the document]
mary-dcouto: done

programs/ziti-edge-tunnel/include/model/dtos.h
```diff
@@ -90,6 +91,11 @@ XX(Duration, int, none, Duration, __VA_ARGS__) \
 XX(StartTime, timestamp, none,StartTime, __VA_ARGS__) \
 XX(Identities, tunnel_identity, array, Identities, __VA_ARGS__)
 
+#define INSTANCE_STATUS(XX, ...) \
+XX(ok, __VA_ARGS__)
```
scareything, line 95: there is only one value in this enum. should it also at least have a "not ok" value? and if so, could it be replaced with a `bool`?

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,158 @@
...
+#define MIN_BUFFER_LEN 512
+
+static uv_mutex_t mutex;
+
+bool initialize_instance_config() {
+    int mutex_init = uv_mutex_init(&mutex);
+    if (mutex_init != 0) {
+        ZITI_LOG(TRACE, "Could not initialize resources for the config, config file will not be updated");
+    }
+    return mutex_init;
```
scareything, line 35: mutex_init is an int, but this function returns bool. furthermore, uv_mutex_init returns 0 on success, which is `false` when interpreted as a boolean.

this function should either return the uv status code directly (as an int), or explicitly map the code to a bool:
```
    return (mutex_init == 0 ? true : false;
```
also, this function's return value is ignored when it is called from `run_tunnel`

programs/ziti-edge-tunnel/instance-config.c
```diff
@@ -0,0 +1,158 @@
...
+}
+
+bool load_tunnel_status_from_file(uv_loop_t* ziti_loop) {
+    char* config_path = get_system_config_path();
+
+    uv_fs_t fs;
+    int check = uv_fs_mkdir(ziti_loop, &fs, config_path, 0755, NULL);
+    if (check == 0) {
+        ZITI_LOG(TRACE, "config path is created at %s", config_path);
+        return false;
```
scareything, line 73: should this return statement be here?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -658,21 +668,43 @@ static void on_event(const base_event *ev) {
...
             controller_event.Identifier = strdup(ev->identifier);
+            controller_event.Fingerprint = strdup(id_event.Id->FingerPrint);
 
             if (zev->code == ZITI_OK) {
                 id_event.Id->Active = true; // determine it from controller
                 if (zev->name) {
-                    id_event.Id->Name = strdup(zev->name);
-                    if (zev->version != NULL) {
+                    if (id_event.Id->Name != NULL && strcmp(id_event.Id->Name, zev->name) != 0) {
+                        if (id_event.Id->Name) free(id_event.Id->Name);
```
scareything, line 684: there is no need to test `id_event.Id->Name` here, as it was tested on the previous line

programs/ziti-edge-tunnel/instance.c
```diff
@@ -34,14 +36,40 @@ tunnel_identity *find_tunnel_identity(const char* identifier) {
...
+            } else {
+                length = strlen(filename);
+            }
+            tnl_id->FingerPrint = calloc(length + 1, sizeof(char));
+            char fingerprint[FILENAME_MAX];
+            memcpy(fingerprint, filename, length);
+            fingerprint[length] = '\0';
+            snprintf(tnl_id->FingerPrint, length+1, "%s", fingerprint);
+
+            if (tnl_id->Name == NULL) {
```
scareything, line 65: Is it necessary to test `tnl_id->Name` for NULL here? I think it will always be NULL, since `tnl_id` was just calloc'ed on line 48.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -658,21 +668,43 @@ static void on_event(const base_event *ev) {
...
+                        if(id_event.Id->ControllerVersion) free(id_event.Id->ControllerVersion);
+                        id_event.Id->ControllerVersion = strdup(zev->version);
+                    } else if (id_event.Id->ControllerVersion == NULL) {
                         id_event.Id->ControllerVersion = strdup(zev->version);
                     }
-                    id_event.Id->Config.ZtAPI = strdup(zev->controller);
+                }
+                if (zev->controller) {
+                    if (id_event.Id->Config != NULL && id_event.Id->Config->ZtAPI != NULL && strcmp(id_event.Id->Config->ZtAPI, zev->controller) != 0) {
+                        if(id_event.Id->Config->ZtAPI) free(id_event.Id->Config->ZtAPI);
```
scareything, line 700: no need to test `id_event.Id->Config->ZtAPI` here

programs/ziti-edge-tunnel/include/model/dtos.h
mary-dcouto: added additional field "initialized" in this enum

programs/ziti-edge-tunnel/instance-config.c
mary-dcouto: fixed the code

programs/ziti-edge-tunnel/include/model/dtos.h
```diff
@@ -92,7 +92,8 @@ XX(StartTime, timestamp, none,StartTime, __VA_ARGS__) \
 XX(Identities, tunnel_identity, array, Identities, __VA_ARGS__)
 
 #define INSTANCE_STATUS(XX, ...) \
-XX(ok, __VA_ARGS__)
+XX(ok, __VA_ARGS__) \
```
scareything, line 95: Sorry to keep coming back to this. I'm just trying to understand why we have an enum here.

1. What is the difference between the "ok" status and the newly added "initialized" status.
2. Why not use a boolean for the "Status" field of the tunnel_identity model instead of this enum?

mary-dcouto: will remove this enum, I thought, if we need to add any other status in future , we could use this enum

mary-dcouto: We fetch the identities records from the config file and then load the identity files. If the identity file is deleted by any chance, then we dont have to send the information of that identity to the UI when the UI is connected. so I added this status field to identify whether the identity file is found in the config location or not. I am making field boolean now

scareything (approved):

PR #231, opened 2021-11-03
Set log level and update tun ipv4/mask

mary-dcouto:


ekoby:
there seems to be a lot of unrelated changes here

more importantly, changing tunnel IP mask requires a full restart because a lot of things would become invalid when it happens on the fly

mary-dcouto:
> there seems to be a lot of unrelated changes here
> 
> more importantly, changing tunnel IP mask requires a full restart because a lot of things would become invalid when it happens on the fly

WDE allows users to change the ip/mask. Once user updates this field, the UI will show a message to restart the tunnel.
I merged the latest main code to this branch, and the base branch of this PR is my 2nd oldest PR, that could be the reason for the unrelated changes. I have raised 2 more PRs before this PR, so if you start reviewing from the oldest PR, that might help.

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
```diff
@@ -128,6 +130,14 @@ XX(down, string, none, down, __VA_ARGS__)
 XX(identifier, string, none, id, __VA_ARGS__) \
 XX(command, TunnelCommand, none, command, __VA_ARGS__)
 
+#define TUNNEL_SET_LOG_LEVEL(XX, ...) \
+XX(loglevel, string, none, Level, __VA_ARGS__)
+
+#define TUNNEL_TUN_IP_V4(XX, ...) \
+XX(tunIP, string, none, TunIPv4, __VA_ARGS__) \
+XX(mask, int, none, TunIPv4Mask, __VA_ARGS__) \
```
scareything, line 138: This field which is called "mask" is actually being used to convey the number of bits in the mask. this is commonly called the "prefix length" .

programs/ziti-edge-tunnel/instance.c
```diff
@@ -313,11 +314,11 @@ static void setTunnelServiceAddress(tunnel_service *tnl_svc, ziti_service *servi
...
             tnl_port_range_arr[port_idx] = getTunnelPortRange(cfg_v1.port_ranges[port_idx]);
         }
-        if (cfg_v1.addresses) free(cfg_v1.addresses);
-        if (cfg_v1.port_ranges) free(cfg_v1.port_ranges);
+        cfg_v1.protocols = NULL;
+        free_ziti_intercept_cfg_v1(&cfg_v1);
     }  else if ((cfg_json = ziti_service_get_raw_config(service, CFG_ZITI_TUNNELER_CLIENT_V1)) != NULL) {
         ZITI_LOG(TRACE, "ziti-tunneler-client.v1: %s", cfg_json);
-        ziti_client_cfg_v1 zt_client_cfg_v1;
+        ziti_client_cfg_v1 zt_client_cfg_v1 = {0};
```
scareything, line 321: initializing with `{ 0 }` is not necessary, since `parse_obj` uses `memset` to zero the object before parsing.

programs/ziti-edge-tunnel/instance.c
```diff
@@ -457,6 +460,7 @@ void set_identifier_from_identities() {
     if (tnl_status.Identities == NULL) {
         return;
     }
+#if _WIN32
```
scareything, line 463: why is this only for WIN32?

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
mary-dcouto: I changed the variable used inside the ZET, but I retained the TunIPv4Mask which gets written into the config file. If we change this now, when we deploy ZET into windows, it will not be able to read this value from the existing TunIPv4Mask value from the config file.

XX(prefixLength, int, none, TunIPv4Mask, __VA_ARGS__) \

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -198,7 +262,10 @@ static void on_cmd(uv_stream_t *s, ssize_t len, const uv_buf_t *b) {
 
         tunnel_comand cmd = {0};
         if (parse_tunnel_comand(&cmd, b->base, len) > 0) {
-            CMD_CTRL->process(&cmd, on_command_resp, s);
+            int status = process_tunnel_commands(&cmd, on_command_resp, s);
```
scareything, line 265: I'm curious, why are some commands handled in `process_tunnel_commands` and others in `process_cmd`?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -177,6 +177,70 @@ static void on_command_resp(const tunnel_result* result, void *ctx) {
...
+            tunnel_tun_ip_v4 tunnel_tun_ip_v4_cmd = {0};
+            if (cmd->data == NULL || parse_tunnel_tun_ip_v4(&tunnel_tun_ip_v4_cmd, cmd->data, strlen(cmd->data)) < 0) {
+                result.error = "invalid command";
+                result.success = false;
+                break;
+            }
+            if (tunnel_tun_ip_v4_cmd.prefixLength < MINTUNIPV4PREFIXLENGTH || tunnel_tun_ip_v4_cmd.prefixLength > MAXTUNIPV4PREFIXLENGTH) {
+                char error_msg[50];
+                snprintf(error_msg, sizeof(error_msg), "prefix length should be between %d and %d", MINTUNIPV4PREFIXLENGTH, MAXTUNIPV4PREFIXLENGTH);
+                result.error = error_msg;
```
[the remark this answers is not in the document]
mary-dcouto: The whole tunnel_result is sent as json to the ipc queue as response

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
mary-dcouto: process_tunnel_commands is used to update the log level and the tun ip information in the config file through IPC command. So when the user restarts the tunnel, the new values will be taken. The config file can be modified only from ziti-edge-tunnel.c file.

process_cmd will delegate the requests to ziti_tunnel_ctrl.c , which is used to perform the operations against the controller. config.file cannot be modified or read from that class

programs/ziti-edge-tunnel/instance.c
```diff
@@ -620,6 +619,24 @@ void set_service_version() {
     tnl_status.ServiceVersion->BuildDate = strdup(ziti_tunneler_build_date());
 }
 
+void set_tun_ipv4_into_instance(char* tun_ip, int mask, bool addDns) {
+    if (tnl_status.TunIpv4 != NULL) free(tnl_status.TunIpv4);
+    tnl_status.TunIpv4 = strdup(tun_ip);
+
+    tnl_status.TunIpv4Mask = mask;
```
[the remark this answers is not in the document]
mary-dcouto: Modified the variable except the one that is used for reading/writing the value to the config file

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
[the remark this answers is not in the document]
scareything: even so, if anyone tried to access `result.error` after this code block it would crash or lead to unexpected results.

Can you point me to the place where result.error is actually used?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
scareything: this would be a great comment in the code.

mary-dcouto: updated the comment

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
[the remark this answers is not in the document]
mary-dcouto: I changed this error statement to 
result.error = "prefix length should be between 10 and 18";

It is used in the callback function - on_command_resp. Here we convert the tunnel_result to json and send it back to the IPC queue
static void on_command_resp(const tunnel_result* result, void *ctx) {
    size_t json_len;
    char *json = tunnel_result_to_json(result, MODEL_JSON_COMPACT, &json_len);

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1684,8 +1800,87 @@ static int get_mfa_codes_opts(int argc, char *argv[]) {
...
-#if _WIN32
+static int set_log_level_opts(int argc, char *argv[]) {
+    static struct option opts[] = {
+            {"loglevel", required_argument, NULL, 'l'},
+    };
+    int c, option_index, errors = 0;
+    optind = 0;
+
+    tunnel_set_log_level *log_level_options = calloc(1, sizeof(tunnel_set_log_level));
+    cmd = calloc(1, sizeof(tunnel_comand));
```
scareything, line 1811: what happens to `cmd`? Unless I'm missing something it seems to be unused and leaked.

scareything: Ahh. I see that it's a global. It's used and freed in `on_connect`.

Is there any way to be sure that on_connect is called exactly once when an option is changed? What prevents on_connect from trying to process the same `cmd` twice in row (e.g. a spurious connect to the pipe)?

mary-dcouto: copied the data into the data parameter in the connect object

scareything (approved):

PR #239, opened 2021-11-22
Ziti dump on exit

mary-dcouto:
ziti dump on exit
get tunnel status
remove identity cmd

I merged main into this branch, so you will find the changes from main branch also in this PR, because the base branch of this PR is my last PR in this repo


lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -496,6 +496,39 @@ static int process_cmd(const tunnel_comand *cmd, command_cb cb, void *ctx) {
             return 0;
         }
 
+        case TunnelCommand_RemoveIdentity: {
+            tunnel_delete_identity delete_id = {0};
```
scareything, line 500: ` = { 0 }` is not needed

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -496,6 +496,39 @@ static int process_cmd(const tunnel_comand *cmd, command_cb cb, void *ctx) {
...
+            if (cmd->data == NULL || parse_tunnel_delete_identity(&delete_id, cmd->data, strlen(cmd->data)) != 0) {
+                result.success = false;
+                result.error = "invalid command";
+                free_tunnel_delete_identity(&delete_id);
+                break;
+            }
+            struct ziti_instance_s *inst = model_map_get(&instances, delete_id.identifier);
+
+            if (inst == NULL) {
```
scareything, line 509: This series of conditionals feels a bit convoluted. I wouldn't mention it but this pattern is repeated a few times.

First the error string and success status are set if `inst == NULL`; after which the `break` on line 513 takes us out of the `TunnelCommand_RemoveIdentity` case.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -496,6 +496,39 @@ static int process_cmd(const tunnel_comand *cmd, command_cb cb, void *ctx) {
...
+                result.error = "ziti context not found";
+                result.success = false;
+                free_tunnel_delete_identity(&delete_id);
+                break;
+            }
+            if (inst) {
```
scareything, line 515: At this point we can be sure that `inst` is not NULL, because the `break` on line 513 would prevent this line from being reached if `inst` was NULL. So at the very least this test on `inst` is not needed here.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -496,6 +496,39 @@ static int process_cmd(const tunnel_comand *cmd, command_cb cb, void *ctx) {
...
+                if (inst->ztx && ziti_get_identity(inst->ztx)) {
+                    disconnect_identity(inst->ztx, CMD_CTX.tunnel_ctx);
+                }
+                model_map_remove(&instances, delete_id.identifier);
+                result.success = true;
+                result.data = tunnel_comand_to_json(cmd, MODEL_JSON_COMPACT, NULL);
+            } else {
```
scareything, line 522: The code in this else is unreachable, again because the `break` on line 513 would jump out of the `TunnelCommand_RemoveIdentity` case if inst was NULL.

The statements in this else are similar to the ones above, so maybe they could be moved there? It is definitely confusing to leave unreachable code lying around.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -496,6 +496,39 @@ static int process_cmd(const tunnel_comand *cmd, command_cb cb, void *ctx) {
...
+                result.success = false;
+                result.error = calloc(1, strlen(delete_id.identifier) + 35);
```
scareything, line 524: when does `error` get processed? will it be freed?

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -496,6 +496,39 @@ static int process_cmd(const tunnel_comand *cmd, command_cb cb, void *ctx) {
...
+                sprintf(result.error, "ziti instance for id %s is not found", delete_id.identifier);
+            }
+
+            free_tunnel_delete_identity(&delete_id);
+            break;
+        }
```
scareything, line 530: overall I'd suggest coding this case with fewer conditionals and breaks. something like this (untested):

```
        case TunnelCommand_RemoveIdentity: {
            tunnel_delete_identity delete_id = { 0 };
            if (cmd->data == NULL || parse_tunnel_delete_identity(&delete_id, cmd->data, strlen(cmd->data)) != 0) {
                result.success = false;
                result.error = "invalid command";
            } else {
                struct ziti_instance_s *inst = model_map_remove(&instances, delete_id.identifier);
                if (inst) {
                    if (inst->ztx && ziti_get_identity(inst->ztx)) {
                        disconnect_identity(inst->ztx, CMD_CTX.tunnel_ctx);
                    }
                    result.success = true;
                    result.data = tunnel_comand_to_json(cmd, MODEL_JSON_COMPACT, NULL);
                } else {
                    result.success = false;
                    result.error = "ziti context not found";
                }

                free_tunnel_delete_identity(&delete_id);
            }
            break;
        }
```

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -171,27 +171,53 @@ static void on_command_resp(const tunnel_result* result, void *ctx) {
     ZITI_LOG(INFO, "resp[%d,len=%zd] = %.*s",
             result->success, json_len, (int)json_len, json, result->data);
 
+    if (result->success && result->data != NULL) {
+        tunnel_comand tnl_res_cmd = {0};
```
scareything, line 175: I know the `tunnel_comand` model was added before this PR, but this is the first time I've noticed "command" is spelled incorrectly.

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
```diff
@@ -68,7 +70,7 @@ XX(identifier, string, none, identifier, __VA_ARGS__)\
 XX(path, string, none, path, __VA_ARGS__)
 
 #define TNL_DISABLE_IDENTITY(XX, ...) \
-XX(path, string, none, path, __VA_ARGS__)
+XX(identifier, string, none, id, __VA_ARGS__)
```
scareything, line 73: `identifier` isn't quite right here, is it? Why not call it "identity"? I also think the json json and C field names should match.

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
```diff
@@ -130,6 +132,9 @@ XX(down, string, none, down, __VA_ARGS__)
 XX(identifier, string, none, id, __VA_ARGS__) \
 XX(command, TunnelCommand, none, command, __VA_ARGS__)
 
+#define TNL_DELETE_IDENTITY(XX, ...) \
+XX(identifier, string, none, identifier, __VA_ARGS__)
```
scareything, line 136: I see, all of these `tnl_*_identlty` models have an `identifier` field. Some of them existed before this PR.

Why does this one use `identifier` for the C field name when the others use `id`? They should at least be consistent for all of the related models.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -225,12 +225,12 @@ static int process_cmd(const tunnel_comand *cmd, command_cb cb, void *ctx) {
...
             if (inst) {
                 disconnect_identity(inst->ztx, CMD_CTX.tunnel_ctx);
-                model_map_remove(&instances, disable_id.path);
+                model_map_remove(&instances, disable_id.identifier);
                 result.success = true;
             } else {
                 result.success = false;
-                result.error = calloc(1, strlen(disable_id.path) + 35);
-                sprintf(result.error, "ziti instance for id %s is not found", disable_id.path);
+                result.error = calloc(1, strlen(disable_id.identifier) + 35);
```
scareything, line 232: will this ever get freed? I think we had a similar issue in a previous PR.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
mary-dcouto: result will be converted to json in the callback function (on_command_resp in ziti-edge-tunnel.c) and the json message will be sent to the IPC queue

mary-dcouto: I will remove this calloc function call

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
mary-dcouto: I changed the id field to identifier in all the models

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
mary-dcouto: removing this calloc

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
mary-dcouto: I changed the id json field to identifier

mary-dcouto: The identity is called as Identifier in the load function, instance map and in all the IPC commands

scareything (approved):

PR #245, opened 2021-12-02
Events ipc command names fix

mary-dcouto:


lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
```diff
@@ -86,39 +87,39 @@ XX(identifier, string, none, id, __VA_ARGS__) \
 XX(dump_path, string, none, dump_path, __VA_ARGS__)
 
 #define TNL_ENABLE_MFA(XX, ...) \
-XX(identifier, string, none, id, __VA_ARGS__)
+XX(identifier, string, none, Identifier, __VA_ARGS__)
```
scareything, line 90: Is there a reason to capitalize the json fields? It seems tedious to me.

scareything: I see. You want to camel-case the fields that are compound words, which is better than using underscores.

programs/ziti-edge-tunnel/netif_driver/windows/tun.c
```diff
@@ -226,7 +224,7 @@ netif_driver tun_open(struct uv_loop_s *loop, uint32_t tun_ip, uint32_t dns_ip,
     return driver;
 }
 
-static int tun_close(struct netif_handle_s *tun) {
```
scareything, line 229: `tun_close` should not be exposed and called directly.

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
mary-dcouto: I made those field names same as what UI currently supports, WDE UI is using CamelCase

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -2252,6 +2280,14 @@ void scm_service_stop() {
 
     remove_all_nrpt_rules();
 
+    // tun close needs to be fixed
```
scareything, line 2283: what's wrong with tun_close? can we fix it now in this PR?  If not, we should create an issue in GH and maybe make this comment a TODO that references the issue.

mary-dcouto: When we call tun_close, the ziti_loop thread is crashing. It looks like it is not killing the sessions gracefully. I will file an issue against this repo for this error.

mary-dcouto: https://github.com/openziti/ziti-tunnel-sdk-c/issues/275

scareything: please add the GitHub issue link as a comment, so it survives this PR.

scareything (approved):

PR #264, opened 2022-01-07
Feature/geneve

dariuszSki:
@ekoby and @scareything just wanted to create this PR , so you can officially review the code. I worked with Shawn quite a bit, so he is familiar with it a bit more already. Please review and provide comments. I have tested the modified code  and seems to be working fine. The throughput is about 60% of the ZET in my testing using the same network and ERs plus load balance for the geneve case. Obvious question is if you want to support multiple drivers. Let me know what you think and how you would see this enabled if accepted. Thanks!


dovholuknf:
@dariuszSki - you found another way of doing this, right? should we keep this PR out here or can we close it?

qrkourier:
@dariuszSki Question for you above from Clint.

dariuszSki:
But my original question was not answered? Support of more than one driver mode?

scareything:
The only reason not to merge this PR as it exists is because it changes the ZET initialization on Linux to use the geneve driver instead of the tun driver.

Adding a special command line option to select between geneve or tun would be a simple but inelegant solution to that problem. Alternatively, we've been talking and thinking about netif_driver extensions for a while now, and I think this would be a good use of the extension mechanism once it exists. Issues #596 and #597 have been created to track the extension mechanism itself and the implementation of this PR as an extension.

PR #271, opened 2022-01-12
enroll unless config file exists (resolves #268)

### LINKED ISSUE #268: ziti-edge-tunnel: enroll should not overwrite json file on failure
right now issuing `ziti-edge-tunnel enroll -j jwt.file -i identity.file` twice will work the first time, then fail the second time. if you make this mistake you MUST recreate your identity file. it would be much better to not overwrite this file if the command is run twice

qrkourier:


qrkourier:
I ran the following command to test the case where the config file already exists and should not be overwritten e.g. `/tmp/tunnelmount/envvar1.json`. I expected to see the error message printed by `ZITI_LOG`, but the process exits 255 without any output at the expected point in execution.

```bash
$ strace /tmp/ziti-edge-tunnel_${ZITI_VERSION} enroll --jwt /tmp/envvar1.jwt --identity /tmp/tunnelmount/envvar1.json &> /tmp/strace-ziti-edge-tunnel.log
```
[strace-ziti-edge-tunnel.log](https://github.com/qrkourier/ziti-tunnel-sdk-c/files/7852256/strace-ziti-edge-tunnel.log)

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1374,16 +1374,21 @@ static void enroll(int argc, char *argv[]) {
...
+            ZITI_LOG(ERROR, "refusing to overwrite existing config file %s: %s(%d)", config_file, strerror(errno), errno);
+            exit(-1);
+        }
+    } else {
+        if ((outfile = fdopen(config_file, "wb")) == NULL) {
+            ZITI_LOG(ERROR, "failed to open file %s: %s(%d)", config_file, strerror(errno), errno);
+            exit(-1);
+        }
     }
-    uv_loop_t *l = uv_loop_new();
```
mary-dcouto, line 1383: Did you remove these 3 lines accidentally?

qrkourier: @mary-dcouto You are correct! I've restored the unintentionally-dropped lines in the last commit.

qrkourier:
@mary-dcouto Thank you for your help. With the build from the change that restored those missing lines I am able to execute normally, but my changes to check for the existing file are not functioning as expected. I will have to try again. Instead of logging a message like "refusing to overwrite existing config file" the file is overwritten.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1374,16 +1374,25 @@ static void enroll(int argc, char *argv[]) {
...
+    #include <errno.h>
     FILE *outfile;
-    if ((outfile = fopen(config_file, "wb")) == NULL) {
-        ZITI_LOG(ERROR, "failed to open file %s: %s(%d)", config_file, strerror(errno), errno);
-        exit(-1);
+    outfile = open(config_file, O_CREAT | O_WRONLY | O_EXCL, S_IRUSR | S_IWUSR);
+    if (outfile < 0) {
+        if (errno == EEXIST) {
+            ZITI_LOG(ERROR, "refusing to overwrite existing config file %s: %s(%d)", config_file, strerror(errno), errno);
+            exit(-1);
```
dovholuknf, line 1384: shouldn't these exits be positive?

qrkourier:
abandoned branch

PR #273, opened 2022-01-13
Tunnel status code

mary-dcouto:
Adding Code field as UI depends on this status code. Changed Status field to IdFileStatus, as old config file already has a Status field which is of string data type

### DIFF
--- a/lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
+++ b/lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
@@ -69,9 +69,10 @@ XX(command, TunnelCommand, none, Function, __VA_ARGS__) \
 XX(data, json, none, Payload, __VA_ARGS__)
 
 #define TUNNEL_CMD_RES(XX, ...) \
-XX(success, bool, none, success, __VA_ARGS__) \
-XX(error, string, none, error, __VA_ARGS__)\
-XX(data, json, none, data, __VA_ARGS__)
+XX(success, bool, none, Success, __VA_ARGS__) \
+XX(error, string, none, Error, __VA_ARGS__)\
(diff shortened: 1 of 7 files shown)


lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
```diff
@@ -307,6 +308,14 @@ struct add_identity_request_s {
     void *cmd_ctx;
 };
 
+#ifndef IPC_SUCCESS
```
scareything, line 311: why protect these in `#ifdef`s? do we want to allow user-supplied values of IPC_SUCCESS / IPC_ERROR?

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
```diff
@@ -307,6 +308,14 @@ struct add_identity_request_s {
...
+#define IPC_SUCCESS 0
+#endif
+#ifndef IPC_ERROR
+#define IPC_ERROR 500
```
scareything, line 315: 500 feels like these are maybe sorta trying to be like http status codes? If so should IPC_SUCCESS be 200?

mary-dcouto: Currently, there are different error codes defined for different types of error. These are not similar to http error codes. However, UI is checking for only 0 (success) and 500(error) error codes, so I copied only those 2 error codes in to t-sdk.

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
mary-dcouto: these are constant fields. I will remove the ifdef

PR #274, opened 2022-01-14
add multi-platform local builder container for Linux

qrkourier:


docker/Dockerfile.linux-build
```diff
@@ -0,0 +1,22 @@
+FROM debian:bullseye-slim
```
scareything, line 1: I'm not sure how much we should care, but this image really assumes an x86_64 docker host. e.g. when building this image on an m1 mac, the x86_64 compiler will not be available.

qrkourier: @scareything Hrmm, true. Ideally the same container-based local build procedure would work for both arm64 and amd64 workstations acting as the Docker host. After I get this working on Linux we can debug any issues that present on your ARM Mac. I assume that Macs that are Docker hosts routinely run a Linux VM to run Linux containers.

qrkourier:
After switching from the generic `cmake --build .`  command to the fancier, arch-specific `cmake` commands I found in the Actions workflow `cmake.yml` I'm now getting this error for both amd64, arm64.

```
[ 71%] Built target mbedcrypto
[ 71%] Built target lwipcore
gmake[2]: Leaving directory '/mnt/build_arm64'
gmake[1]: *** [CMakeFiles/Makefile2:505: CMakeFiles/bundle.dir/rule] Error 2
gmake[1]: Leaving directory '/mnt/build_arm64'
gmake: *** [Makefile:205: bundle] Error 2
```

qrkourier:
Line 505 of Makefile2

```
	$(MAKE) $(MAKESILENT) -f CMakeFiles/Makefile2 CMakeFiles/bundle.dir/all
```

Lines 502-507 of Makefile2

```
# Build rule for subdir invocation for target.
CMakeFiles/bundle.dir/rule: cmake_check_build_system
	$(CMAKE_COMMAND) -E cmake_progress_start /mnt/build-amd64/CMakeFiles 76
	$(MAKE) $(MAKESILENT) -f CMakeFiles/Makefile2 CMakeFiles/bundle.dir/all
	$(CMAKE_COMMAND) -E cmake_progress_start /mnt/build-amd64/CMakeFiles 0
.PHONY : CMakeFiles/bundle.dir/rule
```

scareything:
Yeah, the docker host is always a Linux VM on macOS. The most common docker setup on Mac - Docker Desktop - sets up a vm that is managed by macOS’s built-in xhyve hypervisor. The VM architecture matches the host’s architecture, btw.

qrkourier:
> Yeah, the docker host is always a Linux VM on macOS. The most common docker setup on Mac - Docker Desktop - sets up a vm that is managed by macOS’s built-in xhyve hypervisor. The VM architecture matches the host’s architecture, btw.

Ah, right. So the Linux VM on an M1 Mac will be a Linux arm64 VM. I think this means that `docker run --rm -it -v $PWD:/mnt ziti-edge-tunnel-builder arm64` will work on the Mac, but `docker run --rm -it -v $PWD:/mnt ziti-edge-tunnel-builder amd64` will not because you said the x86_64 compiler won't be available in the arm64 builder container.

README.md
```diff
@@ -124,3 +124,45 @@ Once the Ziti Tunneler SDK is initialized with a network device and ziti-sdk
 callbacks, a tunneler application only needs to indiciate which service(s)
 that should be  
 
+## Multi-Platform Linux Builder Container
```
dovholuknf, line 127: i'd personally like it better if this information was moved to a "BUILDING.md" and then referenced herein. in building.md we'd include simple instructions:
```
mkdir -p build
cd build
cmake .. && cmake --build .
```
for example... and then i also would expect to see the prerequisites covered (not how to install them though) and include all major os mac/linux/windows/ and add in 'docker' here

dovholuknf (approved):

qrkourier:
there's no longer a pressing need for a local cross-build now that we've changed the CI to allow downstream contributors to run the jobs

PR #284, opened 2022-02-01
Remove and add nrpt script

mary-dcouto:
uv_spawn will start a provess asynchronously, so if we call remove nrpt script and then add nrpt script, it is not getting executed in sequence. So I added a remove and add nrpt script function for the change service events

### DIFF
--- a/programs/ziti-edge-tunnel/windows-scripts.c
+++ b/programs/ziti-edge-tunnel/windows-scripts.c
@@ -20,6 +20,7 @@
 #include "windows/windows-scripts.h"
 
 #define MAX_BUCKET_SIZE 512
+#define MIN_BUFFER_LEN 512
 
 static char* const namespace_template = "%s@{n='%s';}";
 static char* const exe_name = "ziti-tunnel";
@@ -199,13 +200,13 @@ void add_nrpt_rules_script(uv_loop_t *nrpt_loop, struct add_or_edit_service_nrpt
     int namespace_template_padding = strlen(namespace_template);
     hostname_list_t host_names_list = LIST_HEAD_INITIALIZER(host_names_list);
     int current_size = 0;
-    int rule_size = 0;
+    int rule_size = MIN_BUFFER_LEN;
     model_map_iter it = model_map_iterator(hostnames);
     while(it != NULL) {
         const char* hostname = model_map_it_key(it);
         if (current_size > MAX_BUCKET_SIZE || rule_size > MAX_POWERSHELL_SCRIPT_LEN) {
             chunked_add_nrpt_rules(nrpt_loop, &host_names_list, dns_ip);
-            rule_size = strlen(hostname) + namespace_template_padding;
+            rule_size = MIN_BUFFER_LEN;
             current_size = 0;
         }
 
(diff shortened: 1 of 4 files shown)


programs/ziti-edge-tunnel/windows-scripts.c
```diff
@@ -362,20 +363,140 @@ void remove_all_nrpt_rules() {
...
+    model_map *hostnames = edit_svc_req_data->hostnames;
+    const char* dns_ip = edit_svc_req_data->dns_ip;
+    if (hostnames == NULL || model_map_size(hostnames) == 0) {
+        ZITI_LOG(DEBUG, "No domains specified to remove_and_add_nrpt_rules, exiting early");
+        return;
+    }
+    int namespace_template_padding = strlen(namespace_template);
+    hostname_list_t host_names_list = LIST_HEAD_INITIALIZER(host_names_list);
+    int current_size = 0;
+    int rule_size = MIN_BUFFER_LEN;
```
scareything, line 469: rule_size should be a `size_t` to capture return values from strlen.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1147,7 +1148,11 @@ static void on_event(const base_event *ev) {
                         for (int i = 0; svc->Addresses[i]; i++) {
                             tunnel_address *addr = svc->Addresses[i];
                             if (addr->IsHost && model_map_get(hostnamesToAdd, addr->HostName) == NULL) {
-                                model_map_set(hostnamesToAdd, addr->HostName, true);
+                                if (model_map_get(hostnamesToRemove, addr->HostName) != NULL) {
+                                    model_map_set(hostnamesToEdit, addr->HostName, true);
```
scareything, line 1152: image

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1147,7 +1148,11 @@ static void on_event(const base_event *ev) {
...
+                                } else {
+                                    model_map_set(hostnamesToAdd, addr->HostName, true);
```
scareything, line 1154: image

It's ok to use the map as a list (of keys), but you can set the values to NULL instead of using an incompatible type for the value.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1157,26 +1162,39 @@ static void on_event(const base_event *ev) {
...
-            if (model_map_size(hostnamesToAdd) > 0 && model_map_size(hostnamesToRemove) > 0) {
-                struct modify_service_nrpt_req *modify_svc_req_data = calloc(1, sizeof(struct modify_service_nrpt_req));
-                modify_svc_req_data->hostnamesToRemove = hostnamesToRemove;
-                modify_svc_req_data->hostnamesToAdd = hostnamesToAdd;
-                modify_svc_req_data->dns_ip = get_dns_ip();
+            // remove the hostnames from hostnamesToRemove, if they are present in hostnamesToEdit
+            if (model_map_size(hostnamesToEdit) > 0) {
+                model_map_iter it = model_map_iterator(hostnamesToRemove);
+                while (it != NULL) {
+                    char *key = model_map_it_key(it);
```
scareything, line 1169: should be const char *

programs/ziti-edge-tunnel/windows-scripts.c
```diff
@@ -362,20 +363,140 @@ void remove_all_nrpt_rules() {
     }
 }
 
+void chunked_remove_and_add_nrpt_rules(uv_loop_t *ziti_loop, hostname_list_t *hostnames, const char* dns_ip) {
+    char* script = calloc(MAX_POWERSHELL_SCRIPT_LEN, sizeof(char));
```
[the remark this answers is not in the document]
mary-dcouto: I tried to change the declaration to char array, but the exec_process is not executing this function. In my IDE I see junk characters, because the memory is not allocated for this script variable properly, when it is passed to the powershell

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
mary-dcouto: I want to check if the value is already added to the map, so I am verifying the map value != NULL in the list 1151 - model_map_get(hostnamesToRemove, addr->HostName) != NULL). So I will use a static string instead of true

programs/ziti-edge-tunnel/windows-scripts.c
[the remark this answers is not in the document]
scareything: How exactly do you declare and initialize the array? If it is being done as in my previous comment and you see garbage characters then I'd suspect an underlying problem with the stack.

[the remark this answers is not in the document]
mary-dcouto: I had seen garbage characters when I ran it in clion once. Now I dont see it anymore. I changed the char pointer to char array.

scareything (approved):

PR #287, opened 2022-02-04
don't clobber Docker registry tags unless forced

qrkourier:


qrkourier:
This is to improve the shell script that I've been using to build the Docker image for ziti-edge-tunnel. Eventually I'll update the build pipeline to create the container images too. I'm waiting to allow some time for us to collect lessons learned from the multi-platform build for ziti-tunnel in the main Ziti repo. When that's well enough burned in then I'll port it over to ziti-edge-tunnel.

docker/buildx.sh
```diff
@@ -88,14 +93,21 @@ fi
...
 
-eval docker buildx build "${DIRNAME}" \
-    --platform="linux/amd64,linux/arm/v7,linux/arm64" \
-    --build-arg=ZITI_VERSION="${ZITI_VERSION}" \
-    "${TAG_PARAMS}" \
-    --push
-
+# if 
+if [[ ${FLAGS:-} =~ f ]] || ! curl -sSLf https://registry.hub.docker.com/v2/repositories/netfoundry/ziti-edge-tunnel/tags/${ZITI_VERSION} &>/dev/null; then
+    eval docker buildx build "${DIRNAME}" \
```
scareything, line 102: why is this being `eval`ed?

qrkourier: @scareything the `eval` is necessary in this case because we're using env var `TAG_PARAMS` to pass command line options to `docker buildx build`. If not eval'd, then quoted, space-separated list of params would be a literal, quoted string as a positional param instead of POSIX arguments and options. It is a choice between not quoting the list (dangerous in a shell script) or using `eval`.

scareything: Alternatively, you could remove the quotation marks that surround the ${TAG_PARAMS} reference.

scareything (approved):

PR #290, opened 2022-02-11
set running event after the adapter creation

mary-dcouto:
when you start the ziti tunnel, it shows the "service is starting" message until the service is set to running state in the services.msc. It was set to running state within a second after starting the ziti-edge-tunnel, now it will be set to running state only after tun is created, which takes around 5-10 seconds


programs/ziti-edge-tunnel/windows-service.c
```diff
@@ -200,14 +201,43 @@ VOID SvcInit( DWORD dwArgc, LPTSTR *lpszArgv)
...
+    }
+
+    // Report running status when initialization is complete.
+
+    while(1)
+    {
+        // If the service receive a running event with in 5 minutes, set the service to running state
+        // otherwise stop the service
+
+        if (WaitForSingleObject(ghSvcRunningEvent, 300000) == WAIT_OBJECT_0) {
```
scareything, line 231: 5 minutes seems like a lot more time that we would want to wait?

scareything (approved):

programs/ziti-edge-tunnel/windows-service.c
mary-dcouto: reduced it to 150 seconds

PR #297, opened 2022-02-23
Fix crash on wake

mary-dcouto:
The control handler for SCM will have to return immediately when it gets a control message. If the function does not return back within 30 seconds, SCM will throw an error. The endpoint status change request must have taken more than 30 seconds when the user resumes the laptop after 30 minutes of sleep, which in turn caused the crash. When the tunneller does not run for more than 30 minutes, the api sessions will expire and the tunneller will become busy to create new api sessions, which causes the endpoint status change function to take longer time to execute.

When the laptop resumes, SCM sends more than 1 event, so the same endpoint status change function will be invoked multiple times. This creates multiple threads, so I added the semaphore lock in the tunnel status get function. It refreshes the global tunnel status instance, which will cause errors if multiple threads try to access at the same time.


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -2600,6 +2596,24 @@ void endpoint_status_change(bool woken, bool unlocked) {
...
+        ZITI_LOG(INFO,"Received session unlocked event");
+    }
+
+    tunnel_status_change *status_change = calloc(1, sizeof(tunnel_status_change));
+    status_change->woken = woken;
+    status_change->unlocked = unlocked;
+
+    uv_async_t *ar = calloc(1, sizeof(uv_async_t));
+    ar->data = status_change;
+    uv_async_init(main_ziti_loop, ar, endpoint_status_change_function);
```
scareything, line 2613: We still have a concurrency issue here, because `uv_async_init` modifies the loop's internal variables (appends to `async_handles`) and cannot be called from multiple threads at the same time. If this was corrected, the tunnel_status semaphore would be unnecessary because `uv_async_send` would ensure that the callback is invoked serially.

Also uv_async_init should (normally) only be called once for a given callback function; otherwise the loop's async_handles field accumulates stale async handles.

My first thought was to add a function to the tSDK that allows an application to request an arbitrary function with arbitrary args to be called on the tunneler_context's loop. But that would require the windows-service code to have a reference to the tunneler_context that was created when ziti_tunneler_init was called. So I'm thinking we either  somehow expose the tunneler_context to ServiceWorkerThread so that it can be passed back to this new "tunneler_call_function" function, or maybe its okay to assume that all tunneler contexts use the same loop.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1201,27 +1211,16 @@ static void on_event(const base_event *ev) {
                 struct add_or_edit_service_nrpt_req *edit_svc_req_data = calloc(1, sizeof(struct add_or_edit_service_nrpt_req));
                 edit_svc_req_data->hostnames = hostnamesToEdit;
                 edit_svc_req_data->dns_ip = get_dns_ip();
-
-                uv_async_t *ar = calloc(1, sizeof(uv_async_t));
-                ar->data = edit_svc_req_data;
-                uv_async_init(main_ziti_loop, ar, remove_and_add_nrpt_rules);
-                uv_async_send(ar);
+                ziti_tunnel_async_send(tunneler, remove_and_add_nrpt_rules, edit_svc_req_data);
```
scareything, line 1214: We should not need to use async calls from within `on_event`, since `on_event` is always called by the ziti_sdk which is on the loop thread.

mary-dcouto: When I called the nrpt_rules scripts directly, ziti tunnel started crashing. So I used this ziti_tunnel_async_send in the on_event functions

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -2600,25 +2640,38 @@ void endpoint_status_change(bool woken, bool unlocked) {
...
+    }
+    if (unlocked) {
+        ZITI_LOG(INFO,"Received session unlocked event");
+    }
+
+    tunnel_status_change *status_change = calloc(1, sizeof(tunnel_status_change));
+    status_change->woken = woken;
+    status_change->unlocked = unlocked;
+
+    ziti_tunnel_async_send(NULL, endpoint_status_change_function, status_change);
```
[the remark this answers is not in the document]
mary-dcouto: endpoint_status_change_function should return immediately, otherwise the scm control handler will fail and the tunnel will crash.

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -603,10 +603,7 @@ static int load_identity(const char *identifier, const char *path, int api_page_
         inst->opts.api_page_size = api_page_size;
     }
 
-    uv_async_t *ar = calloc(1, sizeof(uv_async_t));
-    ar->data = inst;
-    uv_async_init(CMD_CTX.loop, ar, load_ziti_async);
-    uv_async_send(ar);
+    load_ziti_async(CMD_CTX.loop, inst);
```
scareything, line 606: maybe this should be renamed to `load_ziti` since it is being called inline vs through uv_async_send?

lib/ziti-tunnel/ziti_tunnel.c
```diff
@@ -550,6 +559,62 @@ static void run_packet_loop(uv_loop_t *loop, tunneler_context tnlr_ctx) {
...
+    }
+
+    ziti_tunnel_async_call_t *call = calloc(1, sizeof(ziti_tunnel_async_call_t));
+    call->f = f;
+    call->arg = arg;
+
+    uv_async_t *async = calloc(1, sizeof(uv_async_t));
+    async->data = call;
+
+    if (tctx != NULL) {
```
scareything, line 597: could re go back to the version of this function that had pointer variables for the loop and semaphore objects?

mary-dcouto: I added the old code back, please check

scareything (approved):

PR #298, opened 2022-02-25
Kill ziti tunnel on shutdown

mary-dcouto:
As the shutdown request is sent in another thread when the stop event is sent by the SCM, we need to load the wintun library again before deleting it.


programs/ziti-edge-tunnel/netif_driver/windows/tun.c
```diff
@@ -494,3 +494,18 @@ static void cleanup_adapters(wchar_t *tun_name) {
     WintunEnumAdapters(L"Ziti", tun_delete_cb, tun_name);
 }
 
+int tun_kill() {
```
scareything, line 497: I may be missing something, but my first thought is that this should be done from `tun_close`. `tun_close` could also use `ziti_tunnel_async_send` to ensure that everything happens on the loop thread, and then loading the adapter a second time would not be necessary.

mary-dcouto: I changed the code to call the stop operation from the event loop. Now we dont need to load the wintun library twice, but we still need the tun_kill function, because the below line throws segmentation fault if we call the wintun end session after starting the tunnel. I tried to call the tun->close after uv_stop function call, still it seg faults at the same line.
        uv_async_send(tun->read_available);
 So until we fix the run reader issue after session close, we cannot use it. I will file an issue for the same

dovholuknf (approved):

PR #302, opened 2022-02-28
Fix multiple intercept wildcard

ekoby:


lib/ziti-tunnel-cbs/ziti_dns.c
```diff
@@ -233,14 +249,26 @@ static dns_entry_t* new_ipv4_entry(const char *host) {
...
     model_map_set(&ziti_dns.hostnames, host, entry);
-    model_map_set(&ziti_dns.ip_addresses, entry->ip, entry);
+    model_map_set_key(&ziti_dns.ip_addresses, &entry->addr, sizeof(entry->addr), entry);
     ZITI_LOG(INFO, "registered DNS entry %s -> %s", host, entry->ip);
 
     return entry;
 }
 
+const char *ziti_dns_reverse_lookup_domain(const ip_addr_t *addr) {
+    ZITI_LOG(INFO, "matching %s", ipaddr_ntoa(addr));
```
dovholuknf, line 259: this log message seems like it should be at debug not info?

ekoby: yes

dovholuknf (approved):

PR #305, opened 2022-03-07
fix for the semaphore initialization and usage

mary-dcouto:


programs/ziti-edge-tunnel/windows-service.c
```diff
@@ -456,7 +456,7 @@ DWORD LphandlerFunctionEx(
             ReportSvcStatus(SERVICE_STOP_PENDING, NO_ERROR, 0);
 
             // stops the running tunnel service
-            ziti_tunnel_async_send(NULL, scm_service_stop, NULL);
+            scm_service_stop();
```
mary-dcouto, line 459: This function is exposed in the windows-service.h file. If we use the scm_service_stop_event function which accepts uv_loop_t here, we will have to import additional uv libraries in the windows file.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -500,6 +501,23 @@ static bool process_tunnel_commands(const tunnel_command *tnl_cmd, command_cb cb
             free_tunnel_service_control(&tunnel_service_control_opts);
             break;
         }
+        case TunnelCommand_StatusChange: {
```
mary-dcouto, line 504: added this tunnel status command, so we can make this function available from command line. That will help in testing this command without laptop resume operation.

scareything (approved):

PR #306, opened 2022-03-08
Metrics event is so huge and the size of UI logs are in GB. Fixing th…

mary-dcouto:
…e event size

### DIFF
--- a/programs/ziti-edge-tunnel/instance.c
+++ b/programs/ziti-edge-tunnel/instance.c
@@ -424,6 +424,7 @@ tunnel_identity_array get_tunnel_identities_for_metrics() {
         id_new->IdFileStatus = id->IdFileStatus;
         tnl_id_arr[i] = id_new;
     }
+    free(arr);
 
     return tnl_id_arr;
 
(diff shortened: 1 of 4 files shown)


programs/ziti-edge-tunnel/instance.c
```diff
@@ -401,6 +401,34 @@ tunnel_identity_array get_tunnel_identities() {
     return tnl_id_arr;
 }
 
+tunnel_identity_array get_tunnel_identities_for_metrics() {
+
+    tunnel_identity_array arr = get_tunnel_identities();
+    tunnel_identity_array tnl_id_arr = NULL;
```
scareything, line 407: `get_tunnel_identities` never returned NULL. Will callers of this function be prepared for NULL?

scareything (approved):

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -695,7 +695,7 @@ static void send_events_message(const void *message, to_json_fn to_json_f, bool
                 data_len = data_len + sizeof("\n");
```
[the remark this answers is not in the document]
mary-dcouto: changed the code

programs/ziti-edge-tunnel/instance.c
mary-dcouto: yes, the caller is checking for NULL. I added a free(arr) at end of this function now, because get_tunnel_identities() create a new arr array

PR #307, opened 2022-03-09
Add BUILDING.md for Linux

sabedevops:


docs/linux/BUILDING.md
```diff
@@ -0,0 +1,46 @@
...
+scl enable devtoolset-10 bash
+
+cmake -E make_directory ./build
+
+cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=./toolchains/default.cmake -S . -B ./build
+
+cmake --build ./build --target ziti-edge-tunnel --verbose
+
+# Test the built program
+./build/programs/ziti-edge-tunnel/ziti-edge-tunnel version
```
dovholuknf, line 45: might be worthwhile talking about how the container is mounted adn thus how one would be able to pull the build ziti-edge-tunnel out of the docker container

sabedevops: These are not build instructions to a container. This regards only building the project _on_ a CentOS 7 host.

dovholuknf: wow. my bad! how about we change the folder from "linux" then to "centos"? or maybe "linux/centos"?

docs/linux/BUILDING.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Example: Building the `ziti-edge-tunnel` binary:
+
+
+```bash
+#We'll install CMake to this prefix, so let's make sure it's in the PATH
+export PATH="/opt/cmake:${PATH}"
+
+# Install project CMake required version
+curl -L https://cmake.org/files/v3.14/cmake-3.14.0-Linux-x86_64.sh -o ./cmake.sh
```
scareything, line 13: are you getting 3.14 because it is the minimum required version for the tsdk project. If so I think it would be fine to get the latest CMake version that runs on Centos.

docs/linux/BUILDING.md
```diff
@@ -0,0 +1,46 @@
+# CentOS 7:
+
+CentOS 7 ships with older versions of glibc, cmake and gcc that make it difficult to compile this project and its dependencies. Building on CentOS involves providing newer packages built by the upstream projects, or though an available Software Collection Library(SCL).
+
+## Example: Building the `ziti-edge-tunnel` binary:
+
+
+```bash
+#We'll install CMake to this prefix, so let's make sure it's in the PATH
+export PATH="/opt/cmake:${PATH}"
```
scareything, line 10: a variable for "/opt/cmake" wouldn't be a bad idea

scareything (approved):

docs/linux/BUILDING.md
sabedevops: This parrots `openwrt/BUILDING.md` from this repo. It's intended to be expanded via subheadings for other distributions as needed, and is commonly done this way for other projects. Otherwise, we'll be doing `linux/centos/7/BUILDING.md`, `linux/debian/buster/BUILDING.md`. Is that desired?

docs/linux/BUILDING.md
sabedevops: I was indeed using this version for that reason. I've bumped it up to the version sued by the GitHub Actions Ubuntu 18.04 runner.

docs/linux/BUILDING.md
sabedevops: Done.

docs/linux/BUILDING.md
dovholuknf: my only problem is i didn't want to ask you to add the "how to do it for 'non-centos linux'" but... if you want to add that to this .md that's awesome.

i see it as either: make a centos specific building.md - or add the steps for "normal" linux building on here as well (not like that's hard it's probaably a tiny paragraph)

either way works for me

sabedevops: `BUILDING.md ` at the higher level should indeed reflect the full support matrix, including general prerequisites, OS specific dependencies, and build configuration features. As I'm not a project maintainer and would need to thoroughly test the accuracy of that documentation, I've moved to `centos/BUILDING.md` to reflect the more limited scope

scareything: My personal preference would be to frame this doc and others that may follow for different OSs as a "build host setup" doc. Once the host has the right packages, the build process is largely generic as it is driven by cmake.

PR #313, opened 2022-03-12
remove the leading * from an intercept

dovholuknf:


mary-dcouto (approved):

programs/ziti-edge-tunnel/windows-scripts.c
```diff
@@ -97,6 +97,16 @@ static char* exec_process_fetch_result(char* program) {
     return result;
 }
 
+// a function to remove the leading star from a "hostname" when specified from as an intercept
+char *normalize_hostname(char *hostname) {
```
scareything, line 101: the caller should not free or otherwise tamper with the returned string, so returning `const char *` would be a good idea.

scareything (approved):

programs/ziti-edge-tunnel/windows-scripts.c
ekoby: both arg and return should be const

dovholuknf: gah. missed this. will fix

dovholuknf: fixed on desktop-edge-win: https://github.com/openziti/ziti-tunnel-sdk-c/commit/4196803164194a735ce1fd3944528171d104a664

PR #322, opened 2022-03-24
Replaces shell commands with direct systemd-resolved integration

sabedevops:
This PR adds `systemd-resolved` integration via `libsystemd`'s `sd-bus` API. 

A cmake feature flag is included for conditional compilation. Additionally, libsystemd is loaded dynamically at runtime, with fallback to the existing implementation in the case of load failure. 

This approach has some distinct advantages:
* Removes the need to install and run systemd inside container environment. 
    - Only the DBus socket must be shared between the host and container. 
* Detects whether systemd-resolved is registered with DBus, doesn't guess based on list of available binaries
    - Handles systemd installed, but not booted
* Opens the door to respond to signals emitted from DBus in the future. 
    - Example: a route only domain is added to a separate interface, and we want to ensure our interface also sets the route only domain

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -45,12 +45,16 @@ jobs:
       - name: make build dir
         run: cmake -E make_directory ${{runner.workspace}}/build
 
-      - name: install toolchain
+      - name: install crossbuild toolchain
         if: matrix.install != ''
         run: |
           sudo apt update
           sudo apt install ${{ matrix.install }}
 
+      - name: install libsystemd if Ubuntu
+        if: ${{ startsWith(matrix.os, 'ubuntu') }}
+        run: |
+          sudo apt -y install libsystemd-dev
 
       - name: configure cmake
         run: cmake ${{matrix.cmake_opts}} -DCMAKE_TOOLCHAIN_FILE=${{github.workspace}}/toolchains/${{matrix.toolchain}} -S ${{ github.workspace }} -B ${{runner.workspace}}/build
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -54,6 +54,7 @@ jobs:
       - name: install libsystemd if Ubuntu
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         run: |
+          sudo apt update
           sudo apt -y install libsystemd-dev
 
       - name: configure cmake
--- a/programs/ziti-edge-tunnel/netif_driver/linux/resolvers.c
+++ b/programs/ziti-edge-tunnel/netif_driver/linux/resolvers.c
@@ -1,0 +1,424 @@
+/*
+ Copyright 2021 NetFoundry Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#include <net/if.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#ifndef EXCLUDE_LIBSYSTEMD_RESOLVER
+#include <systemd/sd-bus-protocol.h>
+#include <systemd/sd-daemon.h>
+#include <systemd/sd-bus.h>
+#endif
+
+#include <ziti/ziti_log.h>
+#include <uv.h>
+
+#include "utils.h"
+
+#ifndef EXCLUDE_LIBSYSTEMD_RESOLVER
+#ifndef RESOLVED_DBUS_NAME
+#define RESOLVED_DBUS_NAME "org.freedesktop.resolve1"
+#endif
+
+#ifndef RESOLVED_DBUS_PATH
+#define RESOLVED_DBUS_PATH "/org/freedesktop/resolve1"
+#endif
+
+#ifndef RESOLVED_DBUS_MANAGER_INTERFACE
+#define RESOLVED_DBUS_MANAGER_INTERFACE "org.freedesktop.resolve1.Manager"
+#endif
+
+#define _cleanup_(f) __attribute__((cleanup(f)))
+#define TRY_DL(dl_func) do{if ((dl_func) != 0) goto dl_error;} while(0)
+
+static int detect_systemd_resolved_acquired_name(sd_bus *bus);
+static int detect_systemd_resolved_routing_domain_wildcard(sd_bus *bus, int32_t ifindex);
+static bool set_systemd_resolved_link_dns(sd_bus *bus, int32_t ifindex, const char* tun, const char* dns_addr);
+static bool set_systemd_resolved_link_dnssec(sd_bus *bus, int32_t ifindex, const char*tun);
+static bool set_systemd_resolved_link_domain(sd_bus *bus, int32_t ifindex, const char* tun);
+
+// libsystemd prototypes
+static uv_once_t guard;
+static uv_lib_t libsystemd_h;
+static bool libsystemd_dl_success = true;
+static int (*sd_booted_f)(void);
+static int (*sd_bus_call_f)(sd_bus *bus, sd_bus_message *m, uint64_t usec, sd_bus_error *ret_error, sd_bus_message **reply);
+static int (*sd_bus_call_method_f)(sd_bus *bus, const char *destination, const char *path, const char *interface, const char *member, sd_bus_error *ret_error, sd_bus_message **reply, const char *types, ...);
+static void (*sd_bus_error_free_f)(sd_bus_error *e);
+static sd_bus *(*sd_bus_flush_close_unref_f)(sd_bus *bus);
+static int (*sd_bus_get_property_f)(sd_bus *bus, const char *destination, const char *path, const char *interface, const char *member, sd_bus_error *ret_error, sd_bus_message **reply, const char *type);
+static int (*sd_bus_is_bus_client_f)(sd_bus *bus);
+static int (*sd_bus_message_append_f)(sd_bus_message *m, const char *types, ...);
+static int (*sd_bus_message_append_array_f)(sd_bus_message *m, char type, void *ptr, size_t size);
+static int (*sd_bus_message_close_container_f)(sd_bus_message *m);
+static int (*sd_bus_message_enter_container_f)(sd_bus_message *m, char type, const char *contents);
+static int (*sd_bus_message_exit_container_f)(sd_bus_message *m);
+static int (*sd_bus_message_new_method_call_f)(sd_bus *bus, sd_bus_message **m, const char *destination, const char *path, const char *interface, const char *member);
+static int (*sd_bus_message_open_container_f)(sd_bus_message *m, char type, const char *contents);
+static int (*sd_bus_message_read_f)(sd_bus_message *m, const char *types, ...);
+static int (*sd_bus_message_read_strv_f)(sd_bus_message *m, char ***l);
+static sd_bus_message *(*sd_bus_message_unref_f)(sd_bus_message *m);
+static int (*sd_bus_open_system_f)(sd_bus **bus);
+
+static void init_libsystemd() {
+    ZITI_LOG(INFO, "Initializing libsystemd");
+    TRY_DL(uv_dlopen("libsystemd.so.0", &libsystemd_h));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_booted", (void **) &sd_booted_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_call", (void **) &sd_bus_call_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_call_method", (void **) &sd_bus_call_method_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_error_free", (void **) &sd_bus_error_free_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_flush_close_unref", (void **) &sd_bus_flush_close_unref_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_get_property", (void **) &sd_bus_get_property_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_is_bus_client", (void **) &sd_bus_is_bus_client_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_append", (void **) &sd_bus_message_append_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_append_array", (void **) &sd_bus_message_append_array_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_close_container", (void **) &sd_bus_message_close_container_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_enter_container", (void **) &sd_bus_message_enter_container_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_exit_container", (void **) &sd_bus_message_exit_container_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_new_method_call", (void **) &sd_bus_message_new_method_call_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_open_container", (void **) &sd_bus_message_open_container_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_read", (void **) &sd_bus_message_read_f));
+    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_read_strv", (void **) &sd_bus_message_read_strv_f));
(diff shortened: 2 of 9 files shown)


.github/workflows/cmake.yml
```diff
@@ -45,12 +45,17 @@ jobs:
...
         run: cmake -E make_directory ${{runner.workspace}}/build
 
-      - name: install toolchain
+      - name: install crossbuild toolchain
         if: matrix.install != ''
         run: |
           sudo apt update
           sudo apt install ${{ matrix.install }}
 
+      - name: install libsystemd if Ubuntu
```
scareything, line 54: why is this only done on ubuntu? are the libsystemd packages too old on centos? if so I think a comment would be worthwhile here. think about someone adding an OS in the future (e.g. fedora).

scareything (approved):

.github/workflows/cmake.yml
qrkourier: @scareything We could go ahead and handle non-Ubuntu cases here if you prefer that. My first thought was to solve for the OSs we're building at this point. The other PR for packaging does add more OSs, but they're in a separate Actions workflow for now, and so we can solve those package-only builds in a way that makes sense for the way that workflow uses containers.

scareything: I don't think we need to handle other cases here. but I would like to know if/why ubuntu is a special case with respect to libsystemd, and if so, do something here to hopefully avoid ubuntu becoming the only OS that uses libsystemd ever.

scareything: I still don't see why/how libsystemd-dev is only a build-time dependency when building on ubuntu.

qrkourier: @scareything You're wondering why libsystemd-dev is only needed at build-time and not at run-time, correct?

The libsystemd-dev provides the libsystemd header files at build time which are not needed to dynamically-load libsystemd, if present, at runtime in the following new function in `resolvers.c`.

```c
static void init_libsystemd() {
    ZITI_LOG(INFO, "Initializing libsystemd");
    TRY_DL(uv_dlopen("libsystemd.so.0", &libsystemd_h));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_booted", (void **) &sd_booted_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_call", (void **) &sd_bus_call_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_call_method", (void **) &sd_bus_call_method_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_error_free", (void **) &sd_bus_error_free_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_flush_close_unref", (void **) &sd_bus_flush_close_unref_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_get_property", (void **) &sd_bus_get_property_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_is_bus_client", (void **) &sd_bus_is_bus_client_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_append", (void **) &sd_bus_message_append_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_append_array", (void **) &sd_bus_message_append_array_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_close_container", (void **) &sd_bus_message_close_container_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_enter_container", (void **) &sd_bus_message_enter_container_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_exit_container", (void **) &sd_bus_message_exit_container_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_new_method_call", (void **) &sd_bus_message_new_method_call_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_open_container", (void **) &sd_bus_message_open_container_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_read", (void **) &sd_bus_message_read_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_read_strv", (void **) &sd_bus_message_read_strv_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_message_unref", (void **) &sd_bus_message_unref_f));
    TRY_DL(uv_dlsym(&libsystemd_h, "sd_bus_open_system", (void **) &sd_bus_open_system_f));

    goto done;

    dl_error:
    ZITI_LOG(WARN, "Failure during dynamic loading function: %s", uv_dlerror(&libsystemd_h));
    libsystemd_dl_success=false;

    done:
    if (libsystemd_dl_success) {
        ZITI_LOG(DEBUG, "Dynamically loaded libsystemd");
    }
    return;
}
```

[Here's a snip from the x86_64 build of the current PR branch HEAD](https://github.com/openziti/ziti-tunnel-sdk-c/runs/5692734887?check_suite_focus=true). This shows that CMake build option `DISABLE_LIBSYSTEMD_FEATURE` was in the default `OFF` position. This is the part that requires the headers at build time.

```cmake
DISABLE_LIBSYSTEMD_FEATURE: OFF
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1") 
-- Checking for module 'libsystemd'
--   Found libsystemd, version 237
```

This is what the artifact looks like at run-time. Notice that it dynamically loads, not links, libsystemd.so.

```bash
root@34fa455d9045:/# grep -Po '^PRETTY_NAME=\K.*' /etc/os-release
"Ubuntu 18.04.6 LTS"

root@34fa455d9045:/# ldd --verbose ./ziti-edge-tunnel
        linux-vdso.so.1 (0x00007fffffdd3000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3a587de000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3a585bf000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f3a583bb000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f3a581b3000)
        libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f3a57f99000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3a57ba8000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f3a5b0ba000)

        Version information:
        ./ziti-edge-tunnel:
                libm.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libm.so.6
                librt.so.1 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/librt.so.1
                libdl.so.2 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libdl.so.2
                libresolv.so.2 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libresolv.so.2
                libresolv.so.2 (GLIBC_2.9) => /lib/x86_64-linux-gnu/libresolv.so.2
                libpthread.so.0 (GLIBC_2.3.3) => /lib/x86_64-linux-gnu/libpthread.so.0
                libpthread.so.0 (GLIBC_2.3.2) => /lib/x86_64-linux-gnu/libpthread.so.0
                libpthread.so.0 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libpthread.so.0
                libc.so.6 (GLIBC_2.14) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.9) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.25) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.6) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.7) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3.2) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libc.so.6
        /lib/x86_64-linux-gnu/libm.so.6:
                ld-linux-x86-64.so.2 (GLIBC_PRIVATE) => /lib64/ld-linux-x86-64.so.2
                libc.so.6 (GLIBC_2.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_PRIVATE) => /lib/x86_64-linux-gnu/libc.so.6
        /lib/x86_64-linux-gnu/libpthread.so.0:
                ld-linux-x86-64.so.2 (GLIBC_2.2.5) => /lib64/ld-linux-x86-64.so.2
                ld-linux-x86-64.so.2 (GLIBC_PRIVATE) => /lib64/ld-linux-x86-64.so.2
                libc.so.6 (GLIBC_2.14) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3.2) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_PRIVATE) => /lib/x86_64-linux-gnu/libc.so.6
        /lib/x86_64-linux-gnu/libdl.so.2:
                ld-linux-x86-64.so.2 (GLIBC_PRIVATE) => /lib64/ld-linux-x86-64.so.2
                libc.so.6 (GLIBC_PRIVATE) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libc.so.6
        /lib/x86_64-linux-gnu/librt.so.1:
                libpthread.so.0 (GLIBC_2.3.2) => /lib/x86_64-linux-gnu/libpthread.so.0
                libpthread.so.0 (GLIBC_PRIVATE) => /lib/x86_64-linux-gnu/libpthread.so.0
                libpthread.so.0 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libpthread.so.0
                libc.so.6 (GLIBC_2.14) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3.2) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_PRIVATE) => /lib/x86_64-linux-gnu/libc.so.6
        /lib/x86_64-linux-gnu/libresolv.so.2:
                libc.so.6 (GLIBC_2.14) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.4) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_PRIVATE) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.2.5) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.3) => /lib/x86_64-linux-gnu/libc.so.6
        /lib/x86_64-linux-gnu/libc.so.6:
                ld-linux-x86-64.so.2 (GLIBC_2.3) => /lib64/ld-linux-x86-64.so.2
                ld-linux-x86-64.so.2 (GLIBC_PRIVATE) => /lib64/ld-linux-x86-64.so.2

root@34fa455d9045:/# ./ziti-edge-tunnel version --verbose
ziti-tunneler:  v0.17.25-6-g5f86aae
ziti-sdk:       0.26.29

root@34fa455d9045:/# find / -type f -name libsystemd.so*
/lib/x86_64-linux-gnu/libsystemd.so.0.21.0

# libsystemd is already installed
root@34fa455d9045:/# apt-cache show libsystemd0
Package: libsystemd0
Status: install ok installed
Priority: optional
Section: libs
Installed-Size: 657
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: amd64
Multi-Arch: same
Source: systemd
Version: 237-3ubuntu10.53
Pre-Depends: libc6 (>= 2.27), libgcrypt20 (>= 1.8.0), liblz4-1 (>= 0.0~r113), liblzma5 (>= 5.1.1alpha+20120614)
Description: systemd utility library
 The libsystemd0 library provides interfaces to various systemd components.
Description-md5: 92180a6b506aa0eff52136aeb8212dc6
Homepage: https://www.freedesktop.org/wiki/Software/systemd
Original-Maintainer: Debian systemd Maintainers <pkg-systemd-maintainers@lists.alioth.debian.org>
```

Then, the same container when mounted with privileges for the tun device and the dbus socket will dynamically-load libsystemd.so and select systemd as the resolver.

```
root@94ccd7ac151c:/# strace ./ziti-edge-tunnel run 
# ...snip...
[        0.007]    INFO ziti-edge-tunnel:resolvers.c:77 init_libsystemd() Initializing libsystemd
) = 0
epoll_ctl(3, EPOLL_CTL_ADD, 9, {EPOLLIN, {u32=9, u64=9}}) = 0
epoll_ctl(3, EPOLL_CTL_ADD, 12, {EPOLLIN, {u32=12, u64=12}}) = 0
epoll_ctl(3, EPOLL_CTL_ADD, 13, {EPOLLIN, {u32=13, u64=13}}) = 0
epoll_wait(3, [{EPOLLIN, {u32=8, u64=8}}], 1024, 10) = 1
read(8, "\1\0\0\0\0\0\0\0", 1024)       = 8
epoll_wait(3, [        0.007]    INFO ziti-edge-tunnel:resolvers.c:363 try_systemd_resolved() systemd-resolved selected as dns resolver manager
# ...snip...
```

qrkourier: The build-time dep systemd-dev is not installed or needed to run.
```bash
root@94ccd7ac151c:/# apt list --installed 2>/dev/null | grep systemd-dev
root@94ccd7ac151c:/# apt list --installed 2>/dev/null | grep libsystemd0   
libsystemd0/bionic-updates,now 237-3ubuntu10.53 amd64 [installed]
```

scareything: I do understand the build time nature of this dependency. What I don’t understand is why are we only installing it when building on Ubuntu?

sabedevops: The only Linux distribution involved in this particular workflow is Ubuntu 18.04. This differs from the workflow of the CPack PR, which does indeed add `libsystemd` as a build time dependency across all containers that can build with the libsystemd feature.

It is sufficient as written for the purposes of this particular workflow, but I think what you'd like to see is some form of `install libsystemd if linux` in the case additional distributions are added to this workflow in the future? 

An example of how we add it to RHEL8 can be seen here in the CPack PR: 
https://github.com/openziti/ziti-tunnel-sdk-c/pull/321/files#diff-2360cad40a5b33134661683ccc387f197f28844ccd12c9641829ae5382d13a9cR20

scareything: Thanks. I don't expect code changes now (note that I have already approved the PR).

My first comment in this thread still reflects my intent which is to suggest a comment here if the libsystemd requirement on "ubuntu" is only circumstantial.

programs/ziti-edge-tunnel/netif_driver/linux/resolvers.c
```diff
@@ -372,10 +298,23 @@ bool try_systemd_resolved(void) {
...
 
     _cleanup_(sd_bus_flush_close_unrefp_f) sd_bus *bus = NULL;
 
+    // dbus 'ay' encodes 'array of bytes'
+    unsigned char ay[4];
+
+    r = inet_pton(AF_INET, addr, &inaddr);
+
+    if (r != 1) {
+        ZITI_LOG(ERROR, "Failed to translate dns address");
```
scareything, line 312: will it be clear in the logs which dns address couldn't be translated? imo `addr` should be logged with this message.

scareything (approved):

programs/ziti-edge-tunnel/netif_driver/linux/resolvers.c
```diff
@@ -0,0 +1,357 @@
...
+    } else {
+        ZITI_LOG(DEBUG, "Could not detect systemd init system");
+    }
+
+    r = sd_bus_open_system_f(&bus);
+    if ((r >= 0) && (sd_bus_is_bus_client_f(bus) > 0)) {
+        ZITI_LOG(DEBUG, "Connected to system DBus");
+        r = sd_bus_is_acquired_name(bus, RESOLVED_DBUS_NAME);
+        if (r < 0) {
+            ZITI_LOG(ERROR, "Bus name: %s is not an acquired. Falling back to legacy resolvers...", RESOLVED_DBUS_NAME);
```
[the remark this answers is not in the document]
sabedevops: Fixed.

qrkourier (commented):
@sabedevops Are you ready for the final review and merge or have more changes for this PR?

sabedevops:
> @sabedevops Are you ready for the final review and merge or have more changes for this PR?

No more changes planned.

qrkourier:
@scareything Is there a barrier to merging this? Can I help by facilitating the discussion?

qrkourier:
@scareything Thanks, I was able to merge it. This was just confusion and uncertainty on my part because I hadn't merged a PR in the past, and assumed someone else would do so.

PR #325, opened 2022-03-31
Stop service without uv stop

dovholuknf:
get back on the uv loop and clean up. use `uv_cond_wait` to wait until that process completes then exit the main thread after notifying the SCM.

PR #326, opened 2022-04-01
Fix for commandline op

mary-dcouto:


dovholuknf:
replaced with https://github.com/openziti/ziti-tunnel-sdk-c/pull/325

PR #329, opened 2022-04-11
Queue ipc command

mary-dcouto:
I implemented buffering for the ipc commands for windows platform because of the issue that Lars reported. When he tried to add an identity, the jwt content was longer than 1024 chars and the command was split into 2 lines. This issue can occur in other platforms too. In windows, WDE UI sends a '\n' char at the end of the ipc command, so I used that char to determine whether we have received the full ipc command or not. This buffering implementation is done in the similar way as the ziti_queue_work function in the ziti library.


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -48,6 +48,10 @@
 #define MAXMESSAGELEN 4096
 #endif
 
+#ifndef MAXIPCCOMMANDLEN
+#define MAXIPCCOMMANDLEN 4096 * 4
```
dovholuknf, line 52: can't we just put the one if win32 here? are all these other changes actually necessary if we just make the buffer bigger for win32?

dovholuknf: I kinda expected this to be just a single change somewhere - why all the 'extra code'? Also we really need to get away from littering the #if _WIN32 everywhere - there's a lot of them in here ...

mary-dcouto: I moved the MAXIPCCOMMANDLEN macro into the win32 block. I could not find an in built option to increase the buffer length in the uv_read function.

scareything: I also don't see why this shouldn't be a single-line (buffer size) change for all operating systems.

mary-dcouto: I tested the long jwt value in linux, the ipc command didnt get split after 1024 bytes in linux. Probably this buffering issue after 1024 bytes, might be occurring only in windows names pipe, ZDE is not throwing that error in linux when I tested it.

dovholuknf:
@mary-dcouto do you have some easy steps to reproduce this issue? Is it easy to reproduce?

scareything:
These changes were merged with https://github.com/openziti/ziti-tunnel-sdk-c/pull/344

PR #330, opened 2022-04-12
Readme resolver configuration and Dockerfile changes for dbus socket

qrkourier:


README.md
```diff
@@ -57,6 +57,24 @@ RestartSec=3
...
+`ziti-edge-tunnel run` provides a built-in nameserver that will answer queries
+that exactly match authorized OpenZiti services' intercept domain names, and will respond
+with a hard-fail `NXDOMAIN` code if the query does not match an authorized
+service.
+
+Optionally, you may enable DNS recursion by specifying an upstream
+nameserver to answer queries for other domain names that are not services' intercept domain
+names: `ziti-edge-tunnel run --dns-upstream 208.67.222.222`.
+
+`ziti-edge-tunnel` implements the `systemd` dbus socket and will try to use it
```
sabedevops, line 71: uses the `libsystemd` D-Bus RPC client

README.md
```diff
@@ -57,6 +57,24 @@ RestartSec=3
 WantedBy=multi-user.target
 ```
 
+#### How does ziti-edge-tunnel configure nameservers?
+
+`ziti-edge-tunnel run` provides a built-in nameserver that will answer queries
+that exactly match authorized OpenZiti services' intercept domain names, and
+will respond with a hard-fail `NXDOMAIN` code if the query does not match an
```
scareything, line 64: We don't actually return NXDOMAIN. I'm not sure how much detail is needed here, but if you're interested, the bit of our code that handles dns queries is [`process_host_req`](https://github.com/openziti/ziti-tunnel-sdk-c/blob/a33f53370bcca942fe20a6d488c4a55c00b88ca2/lib/ziti-tunnel-cbs/ziti_dns.c#L509):

- looks up the queried hostname in our internal map
- if entry exists, return answer and set query status to `DNS_NO_ERROR`
- queried hostname does not exist in our map, so check if an upstream DNS server is configured and available
- send query to upstream DNS server if one is available
-  if no upstream is available, set query status to `DNS_REFUSE`

scareything (approved):

PR #331, opened 2022-04-12
bugfix 0.17.26: Consume all messages in message loop. 

sabedevops:
@qrkourier Caught this in testing. This happened due to differences in test environment and real system where test environment only had 1 other interface set with the route only domain.


qrkourier (approved):
@scareything I built this branch locally and it solved the problem that I noticed with the build from `main`. WIthout this change I was missing the wildcard search domain on the tunneler device, and so I had found it was necessary without this change to manually configure the search domain for `tun0`.

programs/ziti-edge-tunnel/netif_driver/linux/resolvers.c
```diff
@@ -210,7 +210,6 @@ static int detect_systemd_resolved_routing_domain_wildcard(sd_bus *bus, int32_t
             (domain.is_routing_only == 1)) {
             ZITI_LOG(DEBUG, "systemd-resolved routing only domain wildcard found");
             routing_only_wildcard_set = 0;
-            break;
```
scareything, line 213: It isn't clear to me how this changes the behavior of the code, can you explain why it helps?

If I'm understanding correctly, this loop determines if any interfaces on the host have "." routing domain. If any (even just one) are found then we want to add "." to the routing domains for our tun interface.

The `break` that you removed will only happen when an interface (other than our tun) is found to have "." in its routing domains, at which point we set `routing_only_wildcard_set` to zero to tell the caller that we require "." in our routing domains. Removing the break should only cause the loop to continue through the remaining interfaces - potentially setting `routing_only_wildcard_set` to zero _again_ which would make no difference as far as I can see.

sabedevops: https://www.freedesktop.org/software/systemd/man/sd_bus_message_exit_container

`sd_bus_message_exit_container() exits the scope of the last container entered with sd_bus_message_enter_container(). It behaves mostly the same as sd_bus_message_close_container(). Note that sd_bus_message_exit_container() may only be called after iterating through all members of the container, i.e. reading or skipping them. Use [sd_bus_message_skip(3)](https://www.freedesktop.org/software/systemd/man/sd_bus_message_skip.html#) to skip over fields of a container in order to be able to exit the container with sd_bus_message_exit_container() without reading all members.`

^ This behavior was missed during testing because the containers/vm's I used only had a single additional interface. The code consumed all message container members in that case, when the route-only wildcard was set, so it appeared to work correctly. On a real system, with multiple interfaces, we broke out of the loop without iterating all members of the container. The subsequent call to `sd_bus_message_exit_conainer` therefore failed on the real system due to the behavior described above. Removing break allows us to exhaust all container members, and to end up with the correct direction (yes, sometimes we may set it to found more than once, but we only care that it does get set). The alternative is to add an additional dlsym to get the `sd_bus_message_skip` and to modify the logic to break out of the loop and skip all remaining interfaces. That seemed more complicated than simply allowing it to be found more than once.

sabedevops: Will try these and update the PR.

scareything (approved):

PR #338, opened 2022-04-21
fix posting release artifacts

qrkourier:


.github/workflows/release.yml
```diff
@@ -86,6 +86,9 @@ jobs:
       with:
         path: ${{ runner.workspace }}/downloads/
 
+    - name: List Release Artifacts
```
dovholuknf, line 89: leftover from debugging maybe?

dovholuknf (approved):

scareything (approved):

.github/workflows/release.yml
qrkourier: Intentional in case we need to diagnose a further issue. We could certainly remove it later if there's a reason to do so, but these particular bits only run on a release publish event.

PR #341, opened 2022-04-22
Build updates: EXCLUDE_PROGRAMS, update of catch2 to v2.13.4

smilindave26:


scareything:
Is the catch update needed for an updated apple sdk?

scareything (approved):

smilindave26:
yes - the catch update was needed to build for arm64 on macOS

qrkourier:
It appears that `EXCLUDE_PROGRAMS` and ` TUNNEL_SDK_ONLY` exist for similar purposes. If so, this feels like an opportunity to reduce CMake confusion with comments, at least, if not consolidating these two.

Summary:
* @ekoby added `TUNNEL_SDK_ONLY` to prevent descending into `lib/ziti-tunnel-cbs`
* @smilindave26 added `EXCLUDE_PROGRAMS` to prevent descending into `programs`

Both appear to have the aim of builing only the tunneler SDK without `ziti-edge-tunnel`. Is that accurate?

smilindave26:
TUNNEL_SDK_ONLY also excludes building ZITI-SDK-C (and tunnel_cbs), EXCLUDE_PROGRAM will just skip building the executables.  Yes, a comment would have been a good idea.  This has been merged, so next time somebody is in there adding a comment would be helpful.

ekoby:
> It appears that `EXCLUDE_PROGRAMS` and ` TUNNEL_SDK_ONLY` exist for similar purposes. If so, this feels like an opportunity to reduce CMake confusion with comments, at least, if not consolidating these two.
> 
> Summary:
> * @ekoby added `TUNNEL_SDK_ONLY` to prevent descending into `lib/ziti-tunnel-cbs`
> * @smilindave26 added `EXCLUDE_PROGRAMS` to prevent descending into `programs`
> 
> Both appear to have the aim of builing only the tunneler SDK without `ziti-edge-tunnel`. Is that accurate?

No, TUNNEL_SDK_ONLY is for using tunneler sdk without ziti c-sdk backend. It avoids unnecessary pulling/building of ziti-sdk-c and all of its dependencies

PR #343, opened 2022-04-29
use absolute path to deb conffile

qrkourier:


qrkourier (commented):
test build was installable in https://github.com/openziti/ziti-tunnel-sdk-c/actions/runs/2247577465

scareything (approved):

PR #346, opened 2022-05-03
simplify tun IP configuration

scareything:
- use the base of the dns IP range as the tun IP address
- add 1 (instead of bitwise OR) to the tun IP to get the DNS server listen IP. bitwise OR leads to unexpected values when host portion of base IP is not 0.
- populate IpInfo->IP with the tun IP instead of the DNS IP.

### DIFF
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -1630,9 +1630,7 @@ static void run(int argc, char *argv[]) {
     init = log_init(ziti_loop, multi_writer);
 
     signal(SIGINT, interrupt_handler);
-#endif
 
-#if _WIN32
     if (init) {
         ziti_log_init(ziti_loop, ZITI_LOG_DEFAULT_LEVEL, ziti_log_writer);
         struct tm *start_time = get_log_start_time();
@@ -1679,8 +1677,8 @@ static void run(int argc, char *argv[]) {
         tun_ip |= (ip[i] & 0xFFU);
     }
 
-    tun_ip = htonl(tun_ip);
     uint32_t dns_ip = htonl(tun_ip + 0x1);
+    tun_ip = htonl(tun_ip);
 
 #if __unix__ || __unix
     // prevent termination when running under valgrind
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -1541,7 +1541,7 @@ static struct option run_options[] = {
         { "dns-upstream", required_argument, NULL, 'u'},
 };
 
-static const char* ip_range = "100.64.0.0/10";
+static const char* ip_range = "100.64.0.1/10";
 static const char* dns_impl = NULL;
 static const char* dns_upstream = NULL;
 
@@ -1632,41 +1632,6 @@ static void run(int argc, char *argv[]) {
     signal(SIGINT, interrupt_handler);
 #endif
 
-    char *ip_range_temp = get_ip_range_from_config();
-    if (ip_range_temp != NULL) {
-        ip_range = ip_range_temp;
-    }
-
-    uint32_t ip[4];
-    int bits;
-    int rc = sscanf(ip_range, "%d.%d.%d.%d/%d", &ip[0], &ip[1], &ip[2], &ip[3], &bits);
-    if (rc != 5) {
-        ZITI_LOG(ERROR, "Invalid IP range specification: n.n.n.n/m format is expected");
-        exit(1);
-    }
-
-    uint32_t mask = 0;
-    for (int i = 0; i < 4; i++) {
-        mask <<= 8U;
-        mask |= (ip[i] & 0xFFU);
-    }
-
-    uint32_t tun_ip = htonl(mask | 0x1);
-    uint32_t dns_ip = htonl(mask | 0x2);
-
-#if __unix__ || __unix
-    // prevent termination when running under valgrind
-    // client forcefully closing connection results in SIGPIPE
-    // which causes valgrind to freak out
-    signal(SIGPIPE, SIG_IGN);
-#endif
-
-    // set ip info into instance
-    set_ip_info(dns_ip, tun_ip, bits);
-
-    // set the service version in instance
-    set_service_version();
-
 #if _WIN32
     if (init) {
         ziti_log_init(ziti_loop, ZITI_LOG_DEFAULT_LEVEL, ziti_log_writer);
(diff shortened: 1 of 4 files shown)


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1543,7 +1543,7 @@ static struct option run_options[] = {
         { "dns-upstream", required_argument, NULL, 'u'},
 };
 
-static const char* ip_range = "100.64.0.0/10";
+static const char* ip_range = "100.64.0.1/10";
```
sabedevops, line 1546: Are you sure this works with `ip route add`? I was about to file a separate issue with regards to the present behavior building the main branch. We save `100.64.0.1` to the `/var/lib/ziti/config.json` file into `TunIpv4`. Current main branch build runs successfully the first time, when ip_range is `100.64.0.0/10`, but fails when it reads this value from the config the second time you run it (using TunIpv4 = ip_range = `100.64.0.1/10`. ) with error: 
`Error: Invalid prefix for given prefix length`
image

+cc @dovholuknf

scareything: I've only tested on Windows so far. Thanks for the catch.

I would thin k you see the same problem if you set the dns range to an address with non-zero host bits via `-d`?

I think the linux netif driver needs to make sure that it uses only the network portion of the address when adding the intercept route. I'll update accordingly.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1634,6 +1634,34 @@ static void run(int argc, char *argv[]) {
     signal(SIGINT, interrupt_handler);
 #endif
 
+#if _WIN32
```
dovholuknf, line 1637: nit: there's two #if _WIN32 in a row. seems like this and line 1636 could be removed too, right?

dovholuknf (approved):
lgtm. after updating for running on linux "more than once" :)

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
scareything: yes, the ifdefs can be combined. I reordered this section so the logger was initialized sooner and didn't pay attention to the preceding ifdef at all.

eventually, hopefully soon, all the the OS-specific ifdefs will be removed from ziti-edge-tunnel.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
sabedevops: I just tested this. So it appears we ignore `-d` once the `config.json` file exists. 
image

Trying again after removing the file, it does use the `-d`  flag (until we run it again), and does choke on the `ip_range` as well.
image

scareything: thanks again. IMO we should prioritize the command line options over values in a config file (especially when it is loaded implicitly). I'm not sure if that change will fit in this PR but I'll think about it.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1634,6 +1634,34 @@ static void run(int argc, char *argv[]) {
...
+    ziti_log_init(ziti_loop, ZITI_LOG_DEFAULT_LEVEL, NULL);
+#endif
+
+    // set log level from instance/config, if NULL is returned, the default log level will be used
+    const char* log_lvl = get_log_level();
+    if (log_lvl != NULL) {
+        ziti_log_set_level_by_label(log_lvl);
+    }
+    ziti_tunnel_set_log_level(ziti_log_level());
+    set_log_level(ziti_log_level_label());
```
sabedevops, line 1662: `set_log_level((char *)ziti_log_level_label());`

clang produces a warning without the explicit cast from `const char *` to `char *`:

`passing 'const char *' to parameter of type 'char *' discards qualifiers [clang-diagnostic-incompatible-pointer-types-discards-quali    fiers]`

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
scareything: @sabedevops you'll no longer get the ip route error when using -d (in the absence of a config file)

scareything:
before this change, the value provided with the `-d` option was expected to be a network address (i.e. a CIDR with no host bits set):
- the tun IP was always ".1"
- the dns IP was always ".2"

this PR changes the meaning of the `-d` value:
- the tun IP is the base IP of the provided CIDR.
- the dns IP is the tun IP + ".1".

note that `-d` now works more like the `ip address add` command when adding a CIDR. for example:
```
$ sudo ip addr add 192.168.1.10/24 dev lo
$ ip addr show dev lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet 192.168.1.10/24 scope global lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
```

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
qrkourier: In case others needed to look it up too, `-d` is the abbreviated form of `--dns-ip-range`.

qrkourier:
> before this change, the value provided with the `-d` option was expected to be a network address (i.e. a CIDR with no host bits set):

For reference: `-d` is the abbreviated form of `--dns-ip-range`.

qrkourier:
> this PR changes the meaning of the `-d` value:
> 
> * the tun IP is the base IP of the provided CIDR.
> * the dns IP is the tun IP + ".1".
> 
> note that `-d` now works more like the `ip address add` command when adding a CIDR. for example:

There's an apparent contradiction here if I'm reading you correctly.

If the tun IP (the interface address) is the "base IP of the ... CIDR", then I expect tun IP `192.168.1.0` for DNS IP range `192.168.1.10/24`. However, if the tun IP is assigned in a similar way as `ip address add`, then I expect tun IP `192.168.1.10` for DNS IP range `192.168.1.10/24` (the difference is the host address as network address [base address] of the subnet vs specified host address in IP range).

qrkourier:
@scareything OK, I'll try again to express my understanding in the form of an accurate example.

The default DNS IP range is 100.64.0.1/10 and this will produce a TUN IP of 100.64.0.1 and nameserver IP of 100.64.0.2 because the fourth octet of the specified range is always used as the TUN IP, and the nameserver IP is +1.

If I specify a DNS IP range of 100.64.128.64/26 (64 is the network address) then the TUN IP shall be 100.64.128.65 (network +1) and the nameserver IP 100.64.128.67 (TUN +1).

Just like `ip addr add`, if I specify a DNS IP range where the fourth octet is a usable host address inside the range instead of the network address then the host address becomes the TUN IP instead of computing network address +1. For example, for 100.64.128.68/26 then TUN IP is 100.64.128.68 and nameserver IP is 100.64.128.69.

Is that completely accurate?

scareything:
most of the substantive changes from this PR were added in #470

PR #352, opened 2022-05-11
refine the conditions for sending the Mattermost notification

qrkourier:


.github/workflows/mattermost-ziti-webhook.yml
```diff
@@ -13,13 +13,14 @@ on:
...
     name: POST Webhook
     steps:
-    - uses: openziti/ziti-webhook-action@v1
+    - uses: openziti/ziti-webhook-action@v2
       env:
-        ZITI_MATTERMOST_IDENTITY: ${{ secrets.ZITI_MATTERMOST_IDENTITY }}
-      if: ${{ github.actor!= 'dependabot[bot]' && github.repository_owner == 'openziti' && github.event_name != 'pull_request' && env.ZITI_MATTERMOST_IDENTITY != null }}
+        ZITI_MATTERMOST_WEBHOOK_URL: ${{ secrets.ZITI_MATTERMOST_WEBHOOK_URL }}
+      # conditional filters out noise from the bot and running in forks when the necessary secrets are not configured
+      if: ${{ github.actor!= 'dependabot[bot]' && env.ZITI_MATTERMOST_WEBHOOK_URL != null }}
```
smilindave26, line 23: Should missing env.ZITI_MATTERMOST_WEBHOOK_URL just cause an error (rather than silently skip)?

smilindave26 (approved):

.github/workflows/mattermost-ziti-webhook.yml
smilindave26: ahh - I see.  you are filtering out noise when running from a fork...

qrkourier: @smilindave26 Right, and so for them error would only be an annoyance. The webhook action should fail noisily if any of the three required secrets are missing.

qrkourier: ...if not skipped, I meant to say.

PR #354, opened 2022-05-13
systemd fixes

qrkourier:
@sabedevops 's systemd fixes and a script fix to delete the OTT JWT file after successfully enrolling

### DIFF
--- a/programs/ziti-edge-tunnel/netif_driver/linux/tun.c
+++ b/programs/ziti-edge-tunnel/netif_driver/linux/tun.c
@@ -15,6 +15,7 @@
  */
 
 #include <sys/ioctl.h>
+#include <sys/stat.h>
 #include <sys/wait.h>
 //#include <linux/if.h>
 #include <linux/if_tun.h>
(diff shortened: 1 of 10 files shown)


qrkourier:
### Summary of ZET resolvers logic 

Check whether the libc resolver config file `/etc/resolv.conf` is managed by `systemd-resolved`. We'll know that's the case if it is a symbolic link to one of the four expected configuration files that are provided for this purpose by `systemd-resolved`.

If `/etc/resolv.conf` is managed by `systemd-resolved` then ZET will configure `systemd-resolved` to use and prefer the Ziti nameserver by way of the most modern available interface for `systemd-resolved`. That is, `libsystemd`, the DBUS socket API > `/usr/bin/resolvectl` CLI > `/usr/bin/systemd-resolve` CLI, in that order.

If `/etc/resolv.conf` is not managed by `systemd-resolved` and is a regular file then, we fallback to the legacy resolver CLI  `/usr/sbin/resolvconf` if it is available and valid. We know it's invalid if it is a symbolic link to `systemd-resolved`'s `resolvectl` CLI because at this point in the logic we've already proved that `systemd-resolved` is not managing the resolver config. That link exists on newer systems for which `resolvectl` maintains a reverse-compatible personality when invoked as `resolvconf`.

As a last resort ZET will directly munge `/etc/resolv.conf` to have the Ziti nameserver as primary.

programs/ziti-edge-tunnel/netif_driver/linux/tun.c
```diff
@@ -313,8 +322,8 @@ netif_driver tun_open(uv_loop_t *loop, uint32_t tun_ip, uint32_t dns_ip, const c
     if (driver == NULL) {
         if (error != NULL) {
             snprintf(error, error_len, "failed to allocate netif_device_s");
-            tun_close(tun);
         }
+        tun_close(tun);
```
sabedevops, line 326: Please check me on this change.

programs/ziti-edge-tunnel/netif_driver/linux/tun.c
```diff
@@ -220,7 +229,7 @@ static void init_dns_maintainer(uv_loop_t *loop, const char *tun_name, uint32_t
     if ( s < 0) {
         ZITI_LOG(ERROR, "failed to open netlink socket: %d/%s", errno, strerror(errno));
     }
-    if (bind(s, &local, sizeof(local)) < 0) {
+    if (bind(s, (struct sockaddr *)&local, sizeof(local)) < 0) {
```
sabedevops, line 232: Please check me on this change.

scareything (approved):

PR #357, opened 2022-05-19
C sdk 0 28 2 update in zet

mary-dcouto:


scareything:
I have a branch in progress that adopts the latest ziti SDK. It is nearly done but I've been out this week and unable to complete it.

This PR is not correct as it is. The main issue is that you are using the `addr.hostname` field of the ziti_address model when comparing addresses. `addr.hostname` is not useful when the config contains an IP or CIDR and even when it is a hostname it may contain wildcard domain (with leading asterisk) and should only be used for comparison when `addr.type` is `ziti_address_hostname`. The correct way to compare `ziti_address` values is with the `ziti_address_match` functions.

mary-dcouto:
> I have a branch in progress that adopts the latest ziti SDK. It is nearly done but I've been out this week and unable to complete it.
> 
> This PR is not correct as it is. The main issue is that you are using the `addr.hostname` field of the ziti_address model when comparing addresses. `addr.hostname` is not useful when the config contains an IP or CIDR and even when it is a hostname it may contain wildcard domain (with leading asterisk) and should only be used for comparison when `addr.type` is `ziti_address_hostname`. The correct way to compare `ziti_address` values is with the `ziti_address_match` functions.

If you are already working on it in a different branch, then I can close this one

PR #360, opened 2022-05-23
host only tunneler

mary-dcouto:
adding --disable-intercepts or -h option to the run method that allows user to run the tunneller in host only mode

### DIFF
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -2643,7 +2643,7 @@ static CommandLine run_cmd = make_command("run", "run Ziti tunnel (required supe
                                           " are assigned in N.N.N.N/n format (default 100.64.0.0/10)\n"
                                           "\t-n|--dns <internal|dnsmasq=<dnsmasq opts>> DNS configuration setting (default internal)\n",
         run_opts, run);
-static CommandLine run_host_cmd = make_command("run_host", "run Ziti tunnel to host services",
+static CommandLine run_host_cmd = make_command("run-host", "run Ziti tunnel to host services",
                                           "-i <id.file> [-r N] [-v N]",
                                           "\t-i|--identity <identity>\trun with provided identity file (required)\n"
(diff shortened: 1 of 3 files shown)


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1539,11 +1561,13 @@ static struct option run_options[] = {
         { "refresh", required_argument, NULL, 'r'},
         { "dns-ip-range", required_argument, NULL, 'd'},
         { "dns-upstream", required_argument, NULL, 'u'},
+        { "disable-intercepts", optional_argument, NULL, 'h'},
```
scareything, line 1564: The dns-related arguments will not be used when intercepts are disabled, so it's confusing and misleading for them to be `required_argument`s.

I think the easiest way to avoid potentially conflicting arguments is to add a new top-level command like `run-host` which could accept its own command line arguments if necessary without interfering with the existing `run` arguments.

Another reason I like `run-host` is that it would blend in nicely with a `run-intercept` if we ever wanted to add an intercept-only mode, with `run` continuing to intercept and host services.

mary-dcouto: Created a new top-level command run_host and removed the --disable-intercepts option from the run command. The run function that processes these 2 commands, is not split because it does not require major refactoring apart from 2 blocks of code. Please review and let me know if you want to change that.

scareything (approved):

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -2563,6 +2643,13 @@ static CommandLine run_cmd = make_command("run", "run Ziti tunnel (required supe
                                           " are assigned in N.N.N.N/n format (default 100.64.0.0/10)\n"
                                           "\t-n|--dns <internal|dnsmasq=<dnsmasq opts>> DNS configuration setting (default internal)\n",
         run_opts, run);
+static CommandLine run_host_cmd = make_command("run_host", "run Ziti tunnel to host services",
```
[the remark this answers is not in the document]
mary-dcouto: done

PR #361, opened 2022-05-23
use openssl instead of mbedtls for Linux packages, build a package for ubuntu jammy 22.04 LTS

qrkourier:


qrkourier:
I reverted the change to openssl for RedHat7, Ubuntu14, Ubuntu16 because the build failed on those oldest distros and there seems to be little reason to continually improve those least important packages.

ekoby:
it needs specify openssl(libssl) dependency in the DEB package. Does cmake do it automatically?

qrkourier:
> it needs specify openssl(libssl) dependency in the DEB package. Does cmake do it automatically?

@ekoby Good catch. Fixed.

ekoby:
> > it needs specify openssl(libssl) dependency in the DEB package. Does cmake do it automatically?
> 
> @ekoby Good catch. Fixed.

I don't think we need openssl, just the library package -- `libssl3` seem to be the name

ekoby (approved):

programs/ziti-edge-tunnel/package/CPackGenConfig.cmake
```diff
@@ -29,6 +29,6 @@ if(CPACK_GENERATOR MATCHES "RPM")
     endif(CPACK_GENERATOR MATCHES "RPM")
 
 if(CPACK_GENERATOR MATCHES "DEB")
-    set(CPACK_DEBIAN_PACKAGE_DEPENDS "iproute2, sed, systemd, libatomic1")
+    set(CPACK_DEBIAN_PACKAGE_DEPENDS "iproute2, sed, systemd, libatomic1, openssl")
```
ekoby, line 32: just need the library package `libssl3`

qrkourier: I did find `openssl-libs` for the RPM, but for DEB the `openssl` package has no deps, and so it appears to be the correct package for the runtime. Buildtime uses `libss-dev`.

PR #362, opened 2022-05-24
Disable intercepts in host only mode, supports intercepts only mode

mary-dcouto:
This PR contains the following changes
1. Disable intercepts while running the tunnel in host only mode.
2. New command run_intercepts is added that supports intercepts only mode
3. run certain windows script, only if the user is running the tunnel using admin privileges


qrkourier:
@mary-dcouto We need to eliminate two remaining dependencies when `run-host` mode is invoked:
1. ZET still requires the device /dev/net/tun is present
2. ZET still requires elevated privileges on that device

These need to be eliminated because ZET will never use them at all when using `run-host` mode, and they prevent deploying ZET in a container without special devices and without elevated privileges. This is one of the main goals of the feature, to run in a restricted container with bridge networking where neither of those two requirements can be satisfied.

PR #363, opened 2022-05-25
Service permission in event

mary-dcouto:


scareything:
I think the purpose of this change is to avoid intercept tracking logic when running in host-only mode. Although the logic is superfluous when not intercepting, it is also benign since packets will not be processed at all due to the tun device not being initialized and plugged into the uv loop.

Considering the added complexity of avoiding intercept tacking, I think it's better to just let the logic happen.

PR #376, opened 2022-06-06
removing uv_check and signing off

mary-dcouto:
remove uv_check function


mary-dcouto:
This PR contains the same change as of https://github.com/openziti/ziti-tunnel-sdk-c/pull/367 . I raised this PR, because of the new requirement to signoff all the commits in a branch.

PR #388, opened 2022-06-09
Use ziti_address in intercept and host configurations

scareything:
Update ziti-sdk-c to 0.28.7 and use ziti_address in the service configurations.


.github/workflows/cmake.yml
```diff
@@ -13,27 +13,27 @@ jobs:
...
       matrix:
-        os:
-          - ubuntu-18.04
-          - macOS-10.15
-        cmake_opts: [ "" ]
-        arch: [ x86_64 ]
-        toolchain: [ default.cmake ]
-
         include:
+          - os: macOS-10.15
```
ekoby, line 17: I like this, but add a name to each variant so it is easier to identify in CI `${{ matrix.name }}`

scareything: Thanks. I moved the workflow changes into a separate PR.

PR #391, opened 2022-06-13
Build with visual studio on Windows

scareything:


.github/workflows/cmake.yml
```diff
@@ -53,14 +57,10 @@ jobs:
...
-      - name: install libsystemd if Ubuntu
-        if: ${{ startsWith(matrix.os, 'ubuntu') }}
-        run: |
-          sudo apt -y update
-          sudo apt -y install libsystemd-dev
-
       - name: configure cmake
-        run: cmake ${{matrix.cmake_opts}} -DCMAKE_TOOLCHAIN_FILE=${{github.workspace}}/toolchains/${{matrix.toolchain}} -S ${{ github.workspace }} -B ${{github.workspace}}/build
+        env:
+          TOOLCHAIN: ${{ matrix.toolchain && format('{0}/toolchains/{1}', github.workspace, matrix.toolchain) || '' }}
```
scareything, line 62: for anyone who might be wondering, the trick I'm using to convince `cmake` to use visual studio is to not provide a toolchain spec at all. `cmake` will figure it out.

.github/workflows/cmake.yml
```diff
@@ -53,14 +57,10 @@ jobs:
           sudo apt -y update
           sudo apt -y install ${{ matrix.install }}
 
-      - name: install libsystemd if Ubuntu
```
scareything, line 56: this step isn't necessary if we add "libsystemd-dev to `${{ matrix.install }}`

scareything:
trying to reset checks that are apparently stuck due to "Expected — Waiting for status to be reported"

ekoby (approved):

PR #396, opened 2022-06-20
use string for ziti_dns_setup's dns_addr argument

scareything:
Replace the string argument to ziti_dns_setup. fwiw I originally changed it to ip_addr_t to avoid the int -> string -> int conversion that ziti-edge-tunnel was doing.


smilindave26 (approved):

ekoby (approved):

scareything:
closing and re-opening after updating branch protection rules.

PR #403, opened 2022-06-23
stop requiring docker mount 

qrkourier:
to allow for stuffing identity in local images that are never published

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -29,13 +29,14 @@ function alldone() {
 }
 trap alldone exit
 
-# Ensure that ziti-edge-tunnel's identity is stored on a volume
-# so we don't throw away the one-time enrollment token
-
 IDENTITIES_DIR="/ziti-edge-tunnel"
-if ! mountpoint "${IDENTITIES_DIR}" &>/dev/null; then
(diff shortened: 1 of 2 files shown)


docker/docker-entrypoint.sh
```diff
@@ -33,8 +33,8 @@ trap alldone exit
 # so we don't throw away the one-time enrollment token
 
 IDENTITIES_DIR="/ziti-edge-tunnel"
-if ! mountpoint "${IDENTITIES_DIR}" &>/dev/null; then
-    echo "ERROR: please run this container with a volume mounted on ${IDENTITIES_DIR}" >&2
+if ! [[ -d "${IDENTITIES_DIR}" ]]; then
```
scareything, line 36: The `mountpoint` command was used for this test to try and protect users from storing their json files on an ephemeral volume. I think we are losing that safety net here, which is probably fine - but the comment above should be removed.

scareything (approved):

docker/docker-entrypoint.sh
qrkourier: Comment removed. Warning added.

qrkourier (commented):
I rebased on main following a build check fail

PR #404, opened 2022-06-24
added ability to dial by hostname when using wildecard dns for addressable terminator use where remote identity name is set to fully qualified domain name

r-caamano:


scareything:
this is cool! a couple of thoughts:

1. We are already calculating `dst hostname` in get_app_data_json. it would really be nice if we did the reverse lookup once and passed the result around as needed instead of searching for the IP twice, If you're not up for that right now then maybe we could add an issue to the project for it?
2. I know it's beyond the scope of this PR but  was thinking that the configs could differentiate between the hostname (excluding the wildcard domain) and the fqdn - like $dst_hostname vs $dst_fqdn and maybe even $dst_domain.
3. should the "dst_host" config variable in the dial options match the `dst_hostname` field that we already populate in the app_data?

lib/ziti-tunnel-cbs/ziti_tunnel_cbs.c
```diff
@@ -335,6 +335,9 @@ void * ziti_sdk_c_dial(const void *intercept_ctx, struct io_ctx_s *io) {
             char *proto, *ip, *port;
             strncpy(resolved_dial_identity, dial_opts.identity, sizeof(resolved_dial_identity));
             parse_socket_address(dst_addr, &proto, &ip, &port);
+            const char *dst_host_ip = ip;
+            const char *dst_host = ziti_dns_reverse_lookup(dst_host_ip);
+            printf("\ndest_addr=%s\n", dst_host);
```
scareything, line 340: I think this should be `ZITI_LOG` instead of printf?

scareything (approved):

r-caamano:
closing due to new pull request #414 enhancing the implementation

PR #406, opened 2022-06-24
Refactor appdata

r-caamano:
Refactored fqdn based identity dialing to use appData rather than use reverse lookup and changed variable to $dst_hostname


r-caamano:
Closing seems like we should also pull dest_ip,dst_port and protocol also from app_data

PR #414, opened 2022-06-24
Add ability to dial by fqdn with addressable terminators for use with wildcard DNS

r-caamano:
[refactored refactored resolved_dial_identity in ziti_tunnel_cbs.c to reduce lookups/parcings performed to retrieve dest_xxx variable values as well as add fqdn addressable terminator dialing for use with wildcard DNS.


r-caamano:
closing to add check that &app_data !=NULL before freeing

PR #415, opened 2022-06-25
found issue was that new version was not checking that service_ctx->f…

r-caamano:
…orward_address was true before attempting to see if dst_ip was in service_ctx->addr_u.allowed_addresses. This was causing ip services with out forwardAddress=true to segment fault


scareything:
Thanks for catching this. I see how your change avoids the crash, but I think it would be cleaner to move the dst_ip validation into `addrinfo_from_host_ctx`, in the if/else clause for `dst_ip`.

scareything (approved):

PR #416, opened 2022-06-25
Fixed issue where ip based services with forwardAddress=false cause s…

r-caamano:
Fixed issue issue where ip based services with forwardAddress=false cause set fault.


r-caamano:
made a couple changes

PR #417, opened 2022-06-25
Fixed issue where ip based services with forwardAddress=false cause s…

r-caamano:
Fixed issue where ip based services with forwardAddress=false cause seg fault


ekoby:
it would be (probably) to convert `tunneler_app_data.{dst_ip, dst_address}` field to a `ziti_address`

maybe even combine them into a single field since only one of the values is used and `ziti_address` can cover both options

PR #421, opened 2022-06-27
Implement route update batching

### LINKED ISSUE #418: batch add/delete routes via Netif driver interface
Currently, every route added or removed causes an individual call to add or remove a route.  When disabling an identity, this can lead to a ton of calls.  In macOS, at least, this means a re-config of the tun for every add/remove.  Queuing these and making a single call per identity would be helpful.  This would prob need to add an option delete_route_batch to Netif interface that's only used if present to preserve compatibility (otherwise calling the current add/delete for each route).

ekoby:
API to allow netif drivers to implement route updates batching
implement route batching for Linux

[fixes #418]

### DIFF
--- a/lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
+++ b/lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
@@ -783,6 +783,7 @@ static void on_ziti_event(ziti_context ztx, const ziti_event_t *event) {
                 }
             }
 
+            ziti_tunnel_commit_routes(CMD_CTX.tunnel_ctx);
             break;
         }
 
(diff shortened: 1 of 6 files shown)


lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -783,6 +783,7 @@ static void on_ziti_event(ziti_context ztx, const ziti_event_t *event) {
                 }
             }
 
+            ziti_tunnel_commit_routes(CMD_CTX.tunnel_ctx);
```
smilindave26, line 786: commit always called at end of processing service update - I suppose I could just do this myself at the end of my service update handler.  any merit to the idea of TSDK doing all of the add/remove tracking and sending in a list of routes to add and a list of routes to delete?  That way all TSDK users wouldn't need to implement the logic to keep track of each route add and delete

smilindave26 (approved):

PR #435, opened 2022-07-01
select intercept with closest matching address

### LINKED ISSUE #424: select service with closest matching address when intercepting
The tsdk associates intercepted packets with ziti services by iterating through the available services and selecting the first service with a matching protocol, ip, and port. It would be better to score _all_ service addresses that match the intercepted packet and select the one with the highest score.

For example, two services are being intercepted at the following addresses:
1. tcp:192.168.0.8/32:22
2. tcp:192.168.0.0/24:22

When a packet with dst_ip 192.168.0.8 is intercepted, the connection should be associated with service 1.

scareything:
closes #424

currently in draft status due to untested hostname/wildcard dns intercepts


lib/ziti-tunnel/tests/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+target_link_libraries(ziti-tunnel-sdk-c-test-lib
+        PUBLIC ziti ziti-tunnel-sdk-c
+        )
+
+add_executable(ziti-tunnel-sdk-c-test-runner
+        tunnel_test.cpp
+        )
+
+#target_include_directories(ziti-tunnel-sdk-c-test-runner
```
dovholuknf, line 21: this left here on purpose?

lib/ziti-tunnel/tests/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+#        PUBLIC ${ziti-tunnel-sdk-c_SOURCE_DIR}/tests
+#        PUBLIC ${ziti-tunnel-sdk-c_SOURCE_DIR}/lib/ziti-tunnel
+#        )
+
+target_link_libraries(ziti-tunnel-sdk-c-test-runner
+        PUBLIC ziti-tunnel-sdk-c-test-lib
+        )
+
+if (WIN32)
+    set_property(TARGET ziti-tunnel-sdk-c-test-runner ziti-tunnel-sdk-c-test-lib PROPERTY CXX_STANDARD 20)
```
dovholuknf, line 31: why the difference? also works with mingw and msvc?

lib/ziti-tunnel/include/ziti/ziti_tunnel.h
```diff
@@ -94,9 +94,9 @@ typedef host_ctx_t * (*ziti_sdk_host_cb)(void *ziti_ctx, uv_loop_t *loop, const
 
 /** data needed to intercept packets and dial the associated ziti service */
 typedef struct intercept_ctx_s  intercept_ctx_t;
-typedef bool (*intercept_match_addr_fn)(ip_addr_t *addr, void *app_intercept_ctx);
+typedef const ziti_address * (*intercept_match_addr_fn)(ip_addr_t *addr, void *app_intercept_ctx);
```
dovholuknf, line 97: are these changes breaking changes and thus should get a version bump accordingly?

dovholuknf (approved):

lib/ziti-tunnel/include/ziti/ziti_tunnel.h
scareything: Technically it is a breaking change, but as far as I know this API is only used internally (in ziti-tunnel.lib). I'll increment the version to be sure/clear.

PR #436, opened 2022-07-05
wait N seconds for enroll token or identity file to appear

qrkourier:


docker/docker-entrypoint.sh
```diff
@@ -52,40 +52,55 @@ typeset -a TUNNEL_OPTS
...
+                echo "INFO: trying to get token from stdin" >&2
+                ziti-edge-tunnel enroll --jwt - --identity "${IDENTITY_FILE}" || {
+                    echo "ERROR: failed to enroll with token from stdin" >&2
+                    exit 1
+                }
+            fi
         fi
-    fi
+        # decrement the wait seconds until zero or forever if negative
+        let NF_REG_WAIT--
```
dovholuknf, line 101: so technically this will be "kinda sorta" the number of seconds - which is fine for me. you could use the actual clock time instead to be more accurate if you really care... i doubt we really care though?

dovholuknf (approved):

docker/docker-entrypoint.sh
qrkourier: You're correct. More accurately, it's "at least N seconds", and the realtime will always be a tiny fraction of 1s more than the max N seconds specified.

PR #445, opened 2022-07-15
Allow ZDEW to enumerate all/more/most processes when eval'ing posture checks

### LINKED ISSUE #444: windows cannot enumerate all processes
even when running as administrator or system - the runtime cannot open certain processes without an elevated permission/access token. on startup, set this token when running in windows

dovholuknf:
closes #444 

* remove multiwriter bool and related stuff
* update format away from %FT%T as it is inconsistent on windows. replace with long form
* add function to grant SE_DEBUG to the executing process
* change the way dates print to not be reliant on log init

### DIFF
--- a/programs/ziti-edge-tunnel/windows-service.c
+++ b/programs/ziti-edge-tunnel/windows-service.c
@@ -472,3 +472,42 @@ DWORD LphandlerFunctionEx(
 
     return 0;
 }
+
+
+BOOL SetPrivilege(HANDLE hToken, LPCTSTR lpszPrivilege, BOOL bEnablePrivilege)
+{
+    LUID luid;
+    BOOL bRet=FALSE;
+
+    if (LookupPrivilegeValue(NULL, lpszPrivilege, &luid))
+    {
+        TOKEN_PRIVILEGES tp;
+
+        tp.PrivilegeCount=1;
+        tp.Privileges[0].Luid=luid;
+        tp.Privileges[0].Attributes=(bEnablePrivilege) ? SE_PRIVILEGE_ENABLED: 0;
+        //
+        //  Enable the privilege or disable all privileges.
+        //
+        if (AdjustTokenPrivileges(hToken, FALSE, &tp, NULL, (PTOKEN_PRIVILEGES)NULL, (PDWORD)NULL))
+        {
+            //
+            //  Check to see if you have proper access.
+            //  You may get "ERROR_NOT_ALL_ASSIGNED".
+            //
+            bRet=(GetLastError() == ERROR_SUCCESS);
+        }
+    }
+    return bRet;
+}
+
+void scm_grant_se_debug() {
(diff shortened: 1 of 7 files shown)


programs/ziti-edge-tunnel/windows-service.c
```diff
@@ -472,3 +472,42 @@ DWORD LphandlerFunctionEx(
...
+    return bRet;
+}
+
+void scm_grant_se_debug() {
+    HANDLE hProcess=GetCurrentProcess();
+    HANDLE hToken;
+
+    if (OpenProcessToken(hProcess, TOKEN_ADJUST_PRIVILEGES, &hToken))
+    {
+        SetPrivilege(hToken, SE_DEBUG_NAME, TRUE);
```
scareything, line 510: can this fail?

programs/ziti-edge-tunnel/windows-service.c
```diff
@@ -472,3 +472,42 @@ DWORD LphandlerFunctionEx(
 
     return 0;
 }
+
+
+BOOL SetPrivilege(HANDLE hToken, LPCTSTR lpszPrivilege, BOOL bEnablePrivilege)
```
scareything, line 477: if anything goes wrong in this function it's going to be really hard to tall what happened from the logs

dovholuknf: What would you like to see happen? I'm not sure there's any 'good' way to handle that situation

programs/ziti-edge-tunnel/windows-service.c
dovholuknf: it returns a BOOL - so presumably it can. If it fails, all that should (*should*) happen is that posture checking will fail on more privileged processes. so we should see DEBUG messages in systems that have posture checking enabled and have a process check configured

programs/ziti-edge-tunnel/windows-service.c
scareything: I didn't look into the win32 api to see if the failure cases can be "handled". I'm assuming they can't. I was thinking more about if there's any value to a support engineer or developer in knowing whether LookupPrivilegeValue failed vs AdjustTokenPrivileges? Without knowing the API and reasons for failures it seems like it would be useful to log errors when a function fails, possibly with an error code if available.

programs/ziti-edge-tunnel/windows-service.c
scareything: Might it be nice to also see in the log a WARN (or ERROR) indicating that SetPrivilege failed, and how?

programs/ziti-edge-tunnel/windows-service.c
dovholuknf: Yeah. This runs as basically 'the first command' so logging isn't initialized yet. I suppose it might make sense to move where this happens until after the logging is initialized so that it can be tested and logged - I'll do that...

PR #467, opened 2022-08-05
use the tun's ip for the ip info, not the dns ip

### LINKED ISSUE #466: IpInfo.IP incorrectly set to DNS IP
When saving the config, sending status back to IPC clients etc, this value should be the TUN IP, not the DNS IP

dovholuknf:
closes #466


programs/ziti-edge-tunnel/instance.c
```diff
@@ -683,7 +683,7 @@ void set_ip_info(uint32_t dns_ip, uint32_t tun_ip, int bits) {
     }
     ip_addr_t dns_ip4 = IPADDR4_INIT(dns_ip);
     tnl_status.IpInfo = calloc(1, sizeof(ip_info));
-    tnl_status.IpInfo->Ip = strdup(ipaddr_ntoa(&dns_ip4));
+    tnl_status.IpInfo->Ip = strdup(ipaddr_ntoa(&tun_ip));
```
scareything, line 686: should this be `tun_ip4`?

PR #468, opened 2022-08-05
use info on startup until processing actual log level. fix a warn->debug

### LINKED ISSUE #464: use INFO log level on startup
before processing the log level from the config (windows) or command line, use the default log level of INFO. Switch it to the final level after starting up. This is probably more a problem on windows than other OS's

dovholuknf:
closes #464 

change two warns to debug and stop using `ZITI_LOG_DEFAULT_LEVEL` since the c sdk changed behaviors recently. Use INFO explicitly until processing the actual setting so certain startup messages are not skipped

### DIFF
--- a/lib/ziti-tunnel/ziti_tunnel.c
+++ b/lib/ziti-tunnel/ziti_tunnel.c
@@ -96,7 +96,7 @@ tunneler_context ziti_tunneler_init(tunneler_sdk_options *opts, uv_loop_t *loop)
 
 void ziti_tunnel_commit_routes(tunneler_context tnlr_ctx) {
     if (tnlr_ctx->opts.netif_driver == NULL) {
-        TNL_LOG(WARN, "No netif_driver found tun is running in host only mode and intercepts are disabled");
+        TNL_LOG(DEBUG, "No netif_driver found tun is running in host only mode and intercepts are disabled");
         return;
     }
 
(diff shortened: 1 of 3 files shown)


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1852,8 +1855,6 @@ static void run(int argc, char *argv[]) {
     if (!scm_grant_se_debug()){
         ZITI_LOG(WARN, "could not set se debug access token on process. if process posture checks seem inconsistent this may be why");
     }
-#else
-    ziti_log_init(ziti_loop, ZITI_LOG_DEFAULT_LEVEL, NULL);
```
dovholuknf, line 1856: probably not a big change/diff here... i just moved this up to the previous block where the log init happens to keep the related logic closer

lib/ziti-tunnel/ziti_tunnel.c
```diff
@@ -96,7 +96,7 @@ tunneler_context ziti_tunneler_init(tunneler_sdk_options *opts, uv_loop_t *loop)
 
 void ziti_tunnel_commit_routes(tunneler_context tnlr_ctx) {
     if (tnlr_ctx->opts.netif_driver == NULL) {
-        TNL_LOG(WARN, "No netif_driver found tun is running in host only mode and intercepts are disabled");
+        TNL_LOG(DEBUG, "No netif_driver found tun is running in host only mode and intercepts are disabled");
```
scareything, line 99: I think this is fine and I'm approving, but for the record:

I don't like the original logic here that assumes host-only mode because the netif_driver is null. In a perfect world I'd think the level should be `DEBUG` if host-only and `ERROR` otherwise. Maybe the run mode should be stored in the tunneler context too, so decisions like this are easy to make outside of ziti-edge-tunnel.c

While we're touching this line could the run-on sentence be fixed up?

scareything (approved):

PR #470, opened 2022-08-05
fixes tun ip being incorrectly reported to zdew

### LINKED ISSUE #465: Do not allow Tun IP to be the same as the DNS server
When the IP assigned to the tun is the same as the IP of the DNS service, the tunneler won't be able to intercept the DNS packets and DNS resolution will not succeed.

Decide what to do in this situation and implement the solution. Options include choosing a different DNS address, exiting, logging and moving on, etc

### LINKED ISSUE #466: IpInfo.IP incorrectly set to DNS IP

dovholuknf:
closes #466 
closes #465

* use tun ip for the 'IP' field
* when calculating IP of tun - use the ip provided
* add to the tunip to get the dns ip

### DIFF
--- a/programs/ziti-edge-tunnel/instance.c
+++ b/programs/ziti-edge-tunnel/instance.c
@@ -683,7 +683,7 @@ void set_ip_info(uint32_t dns_ip, uint32_t tun_ip, int bits) {
     }
     ip_addr_t dns_ip4 = IPADDR4_INIT(dns_ip);
     tnl_status.IpInfo = calloc(1, sizeof(ip_info));
-    tnl_status.IpInfo->Ip = strdup(ipaddr_ntoa(&dns_ip4));
+    tnl_status.IpInfo->Ip = strdup(ipaddr_ntoa(&tun_ip4));
     tnl_status.IpInfo->DNS = strdup(ipaddr_ntoa(&dns_ip4));
     tnl_status.IpInfo->MTU = 65535;
 
@@ -691,8 +691,8 @@ void set_ip_info(uint32_t dns_ip, uint32_t tun_ip, int bits) {
     netmask = htonl(netmask);
     ip_addr_t netmask_ipv4 = IPADDR4_INIT(netmask);
     tnl_status.IpInfo->Subnet = strdup(ipaddr_ntoa(&netmask_ipv4));
-
 }
+
 void set_log_level(const char* log_level) {
     if (log_level == NULL) {
         return;
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -124,6 +124,7 @@ static LIST_HEAD(ipc_list, ipc_conn_s) ipc_clients_list = LIST_HEAD_INITIALIZER(
 static long refresh_interval = 10;
 static long refresh_metrics = 5000;
 static long metrics_latency = 5000;
+static char* configured_cidr;
 
 static char *config_dir = NULL;
 
@@ -1663,8 +1664,7 @@ static struct option run_host_options[] = {
         { "refresh", optional_argument, NULL, 'r'},
 };
 
-static const char* ip_range = "100.64.0.0/10";
-static const char* dns_impl = NULL;
+static const char* default_cidr = "100.64.0.1/10";
 static const char* dns_upstream = NULL;
 static bool host_only = false;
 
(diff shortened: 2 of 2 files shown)


programs/ziti-edge-tunnel/instance.c
```diff
@@ -683,16 +683,16 @@ void set_ip_info(uint32_t dns_ip, uint32_t tun_ip, int bits) {
     }
     ip_addr_t dns_ip4 = IPADDR4_INIT(dns_ip);
     tnl_status.IpInfo = calloc(1, sizeof(ip_info));
-    tnl_status.IpInfo->Ip = strdup(ipaddr_ntoa(&dns_ip4));
+    tnl_status.IpInfo->Ip = strdup(ipaddr_ntoa(&tun_ip4));
```
dovholuknf, line 686: this fixes the zdew bug by setting the proper ip into the field

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1663,8 +1664,7 @@ static struct option run_host_options[] = {
         { "refresh", optional_argument, NULL, 'r'},
 };
 
-static const char* ip_range = "100.64.0.0/10";
-static const char* dns_impl = NULL;
+static const char* default_cidr = "100.64.0.1/10";
```
dovholuknf, line 1667: rename "ip_range" to "default_cidr" and change the default to 100.64.0.1 which will become the default IP of the TUN if none are supplied on the command line or via config file

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1818,8 +1823,13 @@ static void run(int argc, char *argv[]) {
...
 
-        tun_ip = htonl(mask | 0x1);
-        dns_ip = htonl(mask | 0x2);
+        tun_ip = htonl(mask);
+        dns_ip = htonl(mask + 1);
+        uint32_t dns_nw_addr = mask & (0xFFFFFFFFUL << (32 - bits)) & 0xFFFFFFFFUL;
+
+        ip_addr_t dns_ip4_addr = IPADDR4_INIT(htonl(dns_nw_addr));
+        dns_range = strdup(ipaddr_ntoa(&dns_ip4_addr));
+        sprintf(dns_range, "%s/%d", strdup(ipaddr_ntoa(&dns_ip4_addr)), bits);
```
dovholuknf, line 1832: make a new char* representing the network address of the "dns range" supplied. this allows one to supply the tun ip and subsequent DNS range in the same field. that's how it works currently but was flawed with the bitwise or'ing going on. instead above we use simple addition. we also do not expect a network address supplied. this is closer to how the zdew works as well.

scareything (approved):

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
scareything: btw the inner `strdup` is unnecessary, and a leak. `ipaddr_ntoa` returns a static buffer, but it's ok because sprintf copies its contents into `dns_range`.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1818,8 +1823,13 @@ static void run(int argc, char *argv[]) {
...
         }
 
-        tun_ip = htonl(mask | 0x1);
-        dns_ip = htonl(mask | 0x2);
+        tun_ip = htonl(mask);
+        dns_ip = htonl(mask + 1);
+        uint32_t dns_nw_addr = mask & (0xFFFFFFFFUL << (32 - bits)) & 0xFFFFFFFFUL;
+
+        ip_addr_t dns_ip4_addr = IPADDR4_INIT(htonl(dns_nw_addr));
+        dns_range = strdup(ipaddr_ntoa(&dns_ip4_addr));
```
scareything, line 1831: `dns_range` needs to be bigger, because you're about to append the prefix to it.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
scareything: alternatively, you could use `ipaddr_ntoa_r` to render the ip string directly into `dns_range` (which still needs to be alloc'ed a little larger), and then sprintf the prefix into `&dns_range[strlen(dns_range)]`.

PR #473, opened 2022-08-09
logging tweaks

scareything:
a few small things that helped when I was debugging voip services.

### DIFF
--- a/lib/ziti-tunnel-cbs/ziti_hosting.c
+++ b/lib/ziti-tunnel-cbs/ziti_hosting.c
@@ -315,8 +315,6 @@ static void on_hosted_udp_server_data(uv_udp_t* handle, ssize_t nread, const uv_
     struct hosted_io_ctx_s *io_ctx = handle->data;
     if (nread > 0) {
         io_ctx->in_wreqs++;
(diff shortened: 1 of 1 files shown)


lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -314,7 +315,9 @@ static void on_hosted_udp_server_data(uv_udp_t* handle, ssize_t nread, const uv_
     struct hosted_io_ctx_s *io_ctx = handle->data;
     if (nread > 0) {
         io_ctx->in_wreqs++;
-        int zs = ziti_write(io_ctx->client, buf->base, nread, on_hosted_ziti_write, buf->base);
+        ZITI_LOG(TRACE, "sending %zd bytes service[%s] client[%s] server[%s]", nread, io_ctx->service->service_name,
```
dovholuknf, line 318: we have users that enable trace logging all the time for troubleshooting. i'd really rather us have some better way of enabling this sort of logging either dynamically or just at startup.

i'd feel better without this TRACE for now? happy to discuss if you want

scareything: as you wish

ekoby (approved):

PR #477, opened 2022-08-10
don't set mfa needed on startup when reading config file

### LINKED ISSUE #476: Tunnel status indicates MFA required after administratively deleting
Right now, when administratively deleting an MFA enrollment, tunnelers will not be notified. When restarting, a tunneler reading the config file, will be marked as MFA required as the value is read from cache. This prevents users from resetting their MFA.

dovholuknf:
closes #476

### DIFF
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -1149,7 +1149,7 @@ static void on_event(const base_event *ev) {
                 controller_event.Fingerprint = strdup(id_event.Id->FingerPrint);
             }
 
-            if (zev->code == ZITI_OK || zev->code == ZITI_PARTIALLY_AUTHENTICATED) {
+            if (zev->code == ZITI_OK) {
                 id_event.Id->Active = true; // determine it from controller
                 if (zev->name) {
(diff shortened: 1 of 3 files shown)


scareything (approved):

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -955,7 +955,9 @@ static notification_message *create_notification_message(tunnel_identity *tnl_id
     notification->Identifier = strdup(tnl_id->Identifier);
     uv_timeval64_t now;
     uv_gettimeofday(&now);
-    notification->MfaTimeDuration = now.tv_sec - tnl_id->MfaLastUpdatedTime->tv_sec;
+    if(tnl_id->MfaLastUpdatedTime) {
```
scareything, line 958: In general, should we test for `now.tv_sec - tnl_id->MfaLastUpdatedTime->tv_sec > 0`?

dovholuknf: I had commented out a line of code around Mfa and this segfaulted. it's not in this pr as I put those lines back but I wasn't expecting a segfault at the place i'd removed the code and it made me nervous that it might happen. That's why I added the check for MfaLastUpdatedTime to be "set" vs NULL

Issue #475, opened 2022-08-09
dns server does not deal with range exhaustion


scareything:
The dns server hands out IPs from the specified DNS range (/10 by default). It does not currently check to see if the range has been exhausted. It should:

1. reset the counter to the beginning of the range when the end of the range is reached
2. never give out an IP that is currently in use (either by a hostname-addressed service, or the dns server IP itself)
3. provide an api for releasing an IP that is no longer in use

PR #478, opened 2022-08-15
recycle DNS ips

scareything:
closes https://github.com/openziti/ziti-tunnel-sdk-c/issues/475

### DIFF
--- a/lib/tests/CMakeLists.txt
+++ b/lib/tests/CMakeLists.txt
@@ -1,0 +1,32 @@
+enable_testing()
+
+# package tests into a library so they can be referenced in all_tests
+add_library(ziti-tunnel-sdk-c-integ-test-lib OBJECT
+        dns_test.cpp
+        )
+
+target_include_directories(ziti-tunnel-sdk-c-integ-test-lib
+        PUBLIC ${ziti-tunnel-sdk-c_SOURCE_DIR}/tests
+        PUBLIC ${ziti-tunnel-sdk-c_SOURCE_DIR}/lib/ziti-tunnel
+        )
+
+target_link_libraries(ziti-tunnel-sdk-c-integ-test-lib
+        PUBLIC ziti-tunnel-sdk-c ziti-tunnel-cbs-c
+        )
+
+add_executable(ziti-tunnel-sdk-c-integ-test-runner
+        tunnel_integ_test.cpp
+        )
+
+target_link_libraries(ziti-tunnel-sdk-c-integ-test-runner
+        PUBLIC ziti-tunnel-sdk-c-integ-test-lib
+        )
+
+if (WIN32)
+    set_property(TARGET ziti-tunnel-sdk-c-integ-test-runner ziti-tunnel-sdk-c-integ-test-lib PROPERTY CXX_STANDARD 20)
+else ()
+    set_property(TARGET ziti-tunnel-sdk-c-integ-test-runner ziti-tunnel-sdk-c-integ-test-lib PROPERTY CXX_STANDARD 11)
+endif ()
+
+include(CTest)
+add_test(quick_tests ziti-tunnel-sdk-c-integ-test-runner -d yes)
--- a/lib/tests/dns_test.cpp
+++ b/lib/tests/dns_test.cpp
@@ -1,0 +1,78 @@
+/*
+ Copyright 2021-2022 NetFoundry Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


lib/tests/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+enable_testing()
+
+# package tests into a library so they can be referenced in all_tests
+add_library(ziti-tunnel-sdk-c-integ-test-lib OBJECT
+        dns_test.cpp
```
ekoby, line 5: does DNS test have to be integ?

scareything: Maybe not, but It was convenient to have a `tunneler_context` when calling `ziti_dns_setup` in the test.

lib/tests/dns_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+    ziti_address_from_string(&za, "just.one.more");
+    CHECK(ziti_dns_register_hostname(&za, new ziti_service) == nullptr);
+
+    // free up an IP and try again
+    ziti_dns_deregister_intercept(model_map_getl(mock_services, 100));
+    ip = ziti_dns_register_hostname(&za, new ziti_service);
+    CHECK(ip != nullptr);
+    CHECK_THAT(ipaddr_ntoa(ip), Catch::Equals("100.64.0.103"));
+
+    // tun ip (.0.1), dns_ip (.0.2), network ip (.0.0), and broadcast ip (.0.255) should not be returned
```
ekoby, line 80: I don't see any other place where network/broadcast IPs are taken into account -- if do intend to take them into account, then the pool capacity should be reduced accordingly

scareything: the loop conditions in `next_ipv4` prevent the last bit in the range from being used, because the candidate is calculated _before_ incrementing the pool counter.

    do {
        candidate = htonl(ziti_dns.ip_pool.base | (ziti_dns.ip_pool.counter++ & ziti_dns.ip_pool.counter_mask));
        i += 1;
        if (ziti_dns.ip_pool.counter == ziti_dns.ip_pool.counter_mask) {
            ziti_dns.ip_pool.counter = 1;
        }
    } while ((model_map_getl(&ziti_dns.ip_addresses, candidate) != NULL) && i < ziti_dns.ip_pool.capacity);

lib/ziti-tunnel-cbs/ziti_dns.c
```diff
@@ -113,7 +114,29 @@ struct ziti_dns_s {
 } ziti_dns;
 
 static uint32_t next_ipv4() {
-   return  htonl(ziti_dns.ip_pool.base | (ziti_dns.ip_pool.counter++ & ziti_dns.ip_pool.counter_mask));
+    uint32_t candidate;
+    uint32_t i = 0; // track how many candidates have been considered. should never exceed pool capacity.
```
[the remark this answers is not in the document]
scareything: The current logic is based on scanning the entire map, because `ziti_dns.ip_addresses` will be fragmented over time. I think we could miss available IPs if we don't check the full range?

The counter in the loop is a fail-safe to avoid looping forever if there is an unforeseen programming error. The test above the loop _should_ prevent us from entering the loop when the scan is doomed.

lib/tests/CMakeLists.txt
scareything: I do think this is the easiest way without some refactoring and exposing `seed_dns` (or a new function that is aware of the tun IP and does all of the dns server setup minus adding the intercept context for the DNS server) to the test.

lib/tests/dns_test.cpp
scareything: Is there anything that needs to be done on this?

I think not, because:
- the loop logic prevents the broadcast IP from being given
- the counter starts at 1, which prevents the network IP from being given
- the pool capacity accounts for the network and broadcast IPs:

      ziti_dns.ip_pool.capacity = (1 << (32 - bits)) - 2; // subtract 2 for network and broadcast IPs

- the tun and dns IPs are reserved by explicitly adding them to the map (with empty dns_entry_t's) in `ziti_dns_setup`

ekoby (approved):

lib/ziti-tunnel-cbs/ziti_dns.c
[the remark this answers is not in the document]
scareything: > would this be easier if we start with capacity - model_map_size(addresses) and just found the next one below?

the ip selection currently depends on incrementing `ip_pool.counter`, and once the counter wraps we don't know if the "next one" is available until we look for it.

PR #492, opened 2022-09-20
Send RST to client when closing un-established connections

### LINKED ISSUE #491: intercepted tcp connections must traverse wait states when ziti_dial fails
We currently close the lwip connection when ziti_dial fails for any reason.

The lwip-side of the client connection is in `SYN_RCVD` until ziti_dial succeeds; so closing the connection (with `tcp_close`) while in this state puts the lwip connection into `FIN_WAIT1`. Now the client is probably confused (rightfully so) and will not send FIN, etc. so the lwip side goes through all of the wait states, and the associated PCB is retained until the timeouts elapse.

scareything:
closes #491

### DIFF
--- a/lib/ziti-tunnel/tunnel_tcp.c
+++ b/lib/ziti-tunnel/tunnel_tcp.c
@@ -236,6 +236,11 @@ int tunneler_tcp_close(struct tcp_pcb *pcb) {
     }
     tcp_arg(pcb, NULL);
     tcp_recv(pcb, NULL);
+    if (pcb->state <= ESTABLISHED) {
+        TNL_LOG(DEBUG, "closing connection before handshake complete. sending RST to client");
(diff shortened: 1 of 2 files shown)


lib/ziti-tunnel/tunnel_tcp.c
```diff
@@ -236,6 +236,11 @@ int tunneler_tcp_close(struct tcp_pcb *pcb) {
     }
     tcp_arg(pcb, NULL);
     tcp_recv(pcb, NULL);
+    if (pcb->state <= ESTABLISHED) {
```
ekoby: should this be `< ESTABLISHED`?

ekoby: looks like it was changed

ekoby (approved):
lgtm

PR #502, opened 2022-10-03
disable systemd feature on OpenWRT builds

ekoby:
script was setting wrong CMAKE option


scareything:
https://github.com/openziti/ziti-tunnel-sdk-c/pull/501

ekoby:
dup

PR #510, opened 2022-10-06
Docker get Ziti ID from env var instead of mounted file

qrkourier:
Also, start using Ziti env var names mapped to existing NF names e.g. `ZITI_ENROLL_TOKEN` instead of `NF_REG_TOKEN` while preserving backwards compat.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -3,15 +3,15 @@ version: "3.9"
 x-base-service: &base-service
     image: openziti/ziti-edge-tunnel
     devices:
-    - /dev/net/tun:/dev/net/tun
+        - /dev/net/tun:/dev/net/tun
     volumes:
-    - .:/ziti-edge-tunnel
-    - /var/run/dbus/system_bus_socket:/var/run/dbus/system_bus_socket
+        - .:/ziti-edge-tunnel
+        - /var/run/dbus/system_bus_socket:/var/run/dbus/system_bus_socket
     environment:
-    - NF_REG_NAME                 # inherit when run like this: NF_REG_NAME=AcmeIdentity docker-compose up ziti-tun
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -3,14 +3,14 @@ version: "3.9"
...
     volumes:
-    - .:/ziti-edge-tunnel
-    - /var/run/dbus/system_bus_socket:/var/run/dbus/system_bus_socket
+        - .:/ziti-edge-tunnel
+        - /var/run/dbus/system_bus_socket:/var/run/dbus/system_bus_socket
     environment:
-    - NF_REG_NAME                 # inherit when run like this: NF_REG_NAME=AcmeIdentity docker-compose up ziti-tun
-    - NF_REG_TOKEN                # NF_REG_NAME=AcmeIdentity NF_REG_TOKEN={JWT} docker-compose up ziti-tun
-    - PFXLOG_NO_JSON=true         # suppress JSON logging
+        - ZITI_IDENTITY_BASENAME  # inherit when run like this: ZITI_IDENTITY_BASENAME=AcmeIdentity docker-compose up ziti-tun
```
scareything, line 11: I'm not sure how the `_BASENAME` suffix helps... Why not `_NAME`?

scareything (approved):

docker/docker-compose.yml
qrkourier: @scareything I chose `_BASENAME` because I have found that it is often confusing for the operator who mistakenly assumes that the correct value here is like `ziti_id.json` instead of the correct value `ziti_id`. I meant to clarify that only the basename of the file, not the full filename, is expected.

NickBenton: I agree, some differentiation is nice here. When I hear _NAME, I think "filename"

PR #515, opened 2022-10-13
add redhat 9 build and package

qrkourier:


.github/actions/openziti-tunnel-build-action/redhat-9/entrypoint.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+#
+# RedHat 8
```
scareything, line 3: RedHat 9

scareything (approved):
I see this is a draft... I'm not sure if that's intentional but fyi it lgtm.

qrkourier:
Draft pending resolution of [related issue](https://github.com/openziti/ziti-tunnel-sdk-c/issues/517#issuecomment-1286197541)

qrkourier:
@scareything Got [a good CPack run](https://github.com/openziti/ziti-tunnel-sdk-c/actions/runs/5215587869) and [successful test enrollment](https://github.com/openziti/ziti-tunnel-sdk-c/issues/517#issuecomment-1583313753)!

scareything:
Nice!

fyi you can link to an issue in a PR just by typing "fixes" and then the issue number preceded by a pound sign. Actually there are [lots of ways] to do this. When thusly linked, the issue is automatically closed when the PR is merged.

scareything:
Can you verify that we're dynamically linking with the ssl libraries from the system?

qrkourier:
> Can you verify that we're dynamically linking with the ssl libraries from the system?

Here are the dynamic links for `ziti-edge-tunnel` I installed in a Rocky Linux 9 LXC container with the RPM I built with the RH9 builder container from this branch.

```bash
[root@rocky9 ~]# ldd /opt/openziti/bin/ziti-edge-tunnel
        linux-vdso.so.1 (0x00007fff83f75000)
        libz.so.1 => /lib64/libz.so.1 (0x00007f15fd370000)
        libatomic.so.1 => /lib64/libatomic.so.1 (0x00007f15fd367000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f15fd28c000)
        libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f15fd278000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f15fd000000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f15fd38f000)
```

scareything:
Is that image used by the cpack workflow?

qrkourier:
> On RHEL9, which resolver works currently on the default OS config? Does RHEL9 enable systemd-resolved by default?

The `rocky9` LXD container and  `generic/rocky9` Vagrant Box for `libvirt` use Network Manager and are not auto-configured by the RPM.

A step is required after installing the RPM to configure Ziti DNS, e.g., running `nmcli connection modify eth0` or an Ansible task:

```yaml
    - name: Add two IPv4 DNS server addresses
      community.general.nmcli:
        conn_name: eth0
        type: ethernet
        dns4_ignore_auto: true
        dns4:
          - 100.64.0.2
          - 1.1.1.1
        state: present
```

qrkourier:
> Is that image used by the cpack workflow?

This branch is up to date with brain `main` and adds a builder image Dockerfile for building the RPM on RedHat9. The `ldd` output I posted is showing the links for the `ziti-edge-tunnel` binary that was included in the RPM built on RedHat9.

scareything:
Ok. We need to figure out why the build isn’t using the system ssl and z libraries. The port overlay cmake variable should cause the builds to link wilt system libraries.

scareything:
ok, I see why the dynamic ssl libraries aren't being linked... duh... it's because we (actually _I_) switched the default back to mbedtls in #663 when we saw the chain verification surprises with openssl recently. Also the cmake variable that controls the crypto libs have changed, and `USE_OPENSSL` no longer gets it done.

so the build for rh9 lgtm.

.github/actions/openziti-tunnel-build-action/redhat-9/entrypoint.sh
```diff
@@ -0,0 +1,56 @@
...
+(
+    [[ -d ./build ]] && rm -r ./build
+    cmake -E make_directory ./build  
+    # allow unset for scl_source scripts
+    set +u
+    cmake \
+        --preset "${cmake_preset}" \
+        -DCMAKE_BUILD_TYPE="${cmake_config}" \
+        -DBUILD_DIST_PACKAGES=ON \
+        -DUSE_OPENSSL=ON \
```
scareything, line 48: as mentioned in the main discussion, this is no longer used and should be removed.

scareything (approved):

PR #524, opened 2022-10-17
fix container terminate

qrkourier:


docker/docker-entrypoint.sh
```diff
@@ -22,12 +22,13 @@ function alldone() {
     # if successfully sent to background then send SIGINT to trigger a cleanup
```
scareything, line 22: change SIGINT to SIGTERM?

PR #525, opened 2022-10-18
build linux artifacts in containers. re-use cmake workflow.

scareything:
Use containers when building linux artifacts. this lets us use any version of any distro that we want for builds, without needing to "upgrade" when GitHub deprecates linux runners.

For now the runner is pinned at ubuntu-20.04 to ensure the oldest possible kernel. I left the build os at ubuntu-18.04, so that the builds are effectively unchanged from what we were getting before this PR. Of course we can now easily go back to an older distro if we want.

The release workflow also re-uses the cmake workflow instead of copying/pasting it. I'm sure we can use similar tricks to fold some of the cpack workflow, but this seems like a worthwhile improvement for now.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -7,9 +7,15 @@ on:
   pull_request:
     branches: [ main ]
 
+  workflow_call:
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+    container: ${{ matrix.container }}
+    env:
+      common_ubuntu_deps: curl doxygen libssl-dev libsystemd-dev pkg-config zlib1g-dev zip
+
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
```diff
@@ -23,63 +29,81 @@ jobs:
...
             arch: x86_64
 
-          - os: ubuntu-latest
+          - os: ubuntu-20.04
             container: ubuntu:18.04
             name: Linux arm
-            install: crossbuild-essential-armhf libsystemd-dev
+            install: $common_ubuntu_deps crossbuild-essential-armhf
             toolchain: Linux-arm.cmake
             cmake_opts: -DCMAKE_BUILD_TYPE=Release
```
scareything, line 43: I noticed that we had not been building our release binaries with openssl. Should we? @ekoby?

ekoby: it is separate discussion

.github/workflows/cmake.yml
```diff
@@ -23,63 +29,81 @@ jobs:
...
+        run: |
+          apt -y update
+          apt -y install software-properties-common
+          add-apt-repository -y ppa:git-core/ppa
+          apt -y update
+          apt -y install git
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+          git --version
+
+      - name: install contemporary cmake
```
ekoby, line 72: there is an action for it: lukka/get-cmake
you get  benefit of caching

ekoby (approved):

PR #530, opened 2022-10-19
Use correct has_arg when parsing command line options with getopt_long

scareything:
`getopt_long` takes an array of `struct option` to represent recognized command line options. `struct option` has a field `has_arg` which specifies whether or not the command line option has an associated argument; possible values are:

- `no_argument`:            no argument to the option is expected
- `required_argument`:      an argument to the option is required
- `optional_argument`:      an argument to the option may be presented

We've been incorrectly setting this field to `optional_argument` in some cases when the intent was to make the command line option itself optional. This PR sets `has_arg` to `required_argument` where appropriate, and adds logic to ensure that required options are passed.


dovholuknf (approved):
lgtm

PR #533, opened 2022-10-19
Don't overwrite identity json when enrolling

### LINKED ISSUE #268: ziti-edge-tunnel: enroll should not overwrite json file on failure

### LINKED ISSUE #516: ziti-edge-tunnel enroll fails silently if identity dir is not writeable
Expected: error message like "ERROR: unable to write identity file in /path/to/identity/dir/config.json, permission denied"
Observed: errexit code 255 is the only clue that enrollment failed

I encountered this in v0.20.0

### LINKED ISSUE #532: enroll command fails unless enroll_key is provided
The fix for #528 added logic to check that required enroll arguments were provided. The logic erroneously checks for the enrollment key instead of the output file.

scareything:
fixes #268
fixes #532
fixes #516

### DIFF
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -2018,6 +2018,10 @@ static void enroll_cb(ziti_config *cfg, int status, char *err, void *ctx) {
 }
 
 static void enroll(int argc, char *argv[]) {
+    uv_loop_t *l = uv_loop_new();
+    int log_level = get_log_level(NULL);
+    ziti_log_init(l, log_level, NULL);
+
     if (config_file == 0) {
         ZITI_LOG(ERROR, "output file option(-i|--identity) is required");
         exit(-1);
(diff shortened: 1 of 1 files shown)


smilindave26 (approved):

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -2018,6 +2025,10 @@ static void enroll_cb(ziti_config *cfg, int status, char *err, void *ctx) {
 }
 
 static void enroll(int argc, char *argv[]) {
+    uv_loop_t *l = uv_loop_new();
```
dovholuknf, line 2028: `process_tunnel_commands` calls `enroll_ziti_async` which calls this function. is it appropriate to get the log level and log init here?

dovholuknf: also seems like `enroll` should take the loop?

dovholuknf: nevermind. it uses `ziti_enroll` instead. my bad! :) lgtm

dovholuknf (approved):

PR #538, opened 2022-10-24
Ensure that BYTE_ORDER is set

scareything:
The `BYTE_ORDER` C preprocessor macro is needed by lwip, but the lwip includes (and -contrib packages) don't always include the correct system includes to pick up a definition of `BYTE_ORDER`. It turns out that there is no standard include file for getting `BYTE_ORDER`, so this PR relies on a `cmake` variable, but only if `BYTE_ORDER` is not already defined.

### DIFF
--- a/lib/ziti-tunnel/CMakeLists.txt
+++ b/lib/ziti-tunnel/CMakeLists.txt
@@ -35,7 +35,7 @@ set (LWIP_INCLUDE_DIRS
 include(${LWIP_DIR}/src/Filelists.cmake)
 
 target_sources(lwipcore PRIVATE ${lwip_sys_srcs} lwip/lwiphooks_ip6.c lwip/lwiphooks_ip4.c lwip/lwip_cloned_fns.c)
-target_compile_definitions(lwipcore PUBLIC CMAKE_C_BYTE_ORDER=${CMAKE_C_BYTE_ORDER})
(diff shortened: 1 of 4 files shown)


scareything:
closes #534

dovholuknf (approved):

lib/ziti-tunnel/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ set (LWIP_INCLUDE_DIRS
 include(${LWIP_DIR}/src/Filelists.cmake)
 
 target_sources(lwipcore PRIVATE ${lwip_sys_srcs} lwip/lwiphooks_ip6.c lwip/lwiphooks_ip4.c lwip/lwip_cloned_fns.c)
+target_compile_definitions(lwipcore PUBLIC CMAKE_C_BYTE_ORDER=${CMAKE_C_BYTE_ORDER})
```
ekoby, line 38: I would not name it CMAKE_xxx

LWIP_BYTE_ORDER would be better

ekoby: why not just set BYTE_ORDER directly here?

scareything: Mainly because the value of the cmake variable is either `BIG_ENDIAN` or `LITTLE_ENDIAN`, so we need to make sure those macros are defined as well but cmake doesn't provide them.

PR #544, opened 2022-11-04
crossbuild arm package

### LINKED ISSUE #449: build broken for bundle target for ARM architectures when USE_OPENSSL=ON
I tried to build the arm64 binary and ran into an issue with USE_OPENSSL=ON. I recognize that we're using mbedtls in GitHub CI, and the package builds are amd64 only, and so I'm wondering if building for ARM architectures with openssl is supported at all. I've attached the build log: [arm64-openssl.log](https://github.com/openziti/ziti-tunnel-sdk-c/files/9152682/arm64-openssl.log)

Here's the incantation I attempted to build for ARM64.

```bash
cmake \
                    -DCMAKE_BUILD_TYPE=Release \
                    -DCMAKE_TOOLCHAIN_FILE=${REPO_DIR}/toolchains/Linux-arm64.cmake \
                    -DUSE_OPENSSL=ON \
                    -S ${REPO_DIR} \
                    -B ${CMAKE_BUILD_DIR} \
&& cmake \
                    --build ${CMAKE_BUILD_DIR} \
                    --target bundle \
                    --verbose;
```

qrkourier:
resolves #449


.github/workflows/cpack.yml
```diff
@@ -15,7 +15,9 @@ jobs:
   package:
     runs-on: ubuntu-20.04
     # optionally override entire container image:tag string
-    container: ${{ matrix.distro.container || format('{0}:{1}', matrix.distro.name, matrix.distro.version) }} 
+    container: 
+      image: ${{ matrix.distro.container || format('{0}:{1}', matrix.distro.name, matrix.distro.version) }}
+      options: --platform=${{ matrix.arch.cmake }}
```
scareything, line 20: the platform argument must match the format `os[/arch[/variant]]`. e.g. `linux/aarch64`

https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerCreate

qrkourier: I haven't yet found a value other than `amd64` and `linux/amd64` that works for `--platform`. The symptom when it's not working is the container stops running before the first job step runs.

.github/actions/openziti-tunnel-build-action/ubuntu-22.04/Dockerfile
```diff
@@ -1,6 +1,11 @@
-FROM ubuntu:jammy
+ARG CMAKE_VERSION="3.22.3"
 
-LABEL org.opencontainers.image.authors="netfoundry.io"
+# upstream of Ubuntu Jammy 22.04 LTS
+FROM debian:bookworm
```
qrkourier, line 4: Debian Bookworm is the contemporary of Ubuntu Jammy, but there's always the possibility that Bookworm will roll forward to a new GLIBC that is unavailable in Jammy. We have the option now, or later, to switch Jammy's build to the predecessor, Debian Bullseye, which will always have a compatible GLIBC, but may introduce other OS differences.

.github/actions/openziti-tunnel-build-action/ubuntu-20.04/Dockerfile
```diff
@@ -1,10 +1,11 @@
 ARG CMAKE_VERSION="3.22.3"
 
-FROM ubuntu:focal
+# upstream of Ubuntu Focal 20.04 LTS
+FROM debian:bullseye
```
qrkourier, line 4: Debian Bullseye is the contemporary of Ubuntu Focal, but there's always the possibility that Bullseye will roll forward to a new GLIBC that is unavailable in Focal. We have the option now, or later, to switch Focal's build to the predecessor, Debian Buster, which will always have a compatible GLIBC, but may introduce other OS differences.

scareything: what's the advantage of change the base image? it sounds like ubuntu is the safer bet?

.github/actions/openziti-tunnel-build-action/ubuntu-20.04/Dockerfile
```diff
@@ -28,9 +34,16 @@ RUN apt-get -y update \
         libssl-dev \
     && rm -rf /var/lib/apt/lists/*
 
-RUN curl -L https://cmake.org/files/v${CMAKE_VERSION%.*}/cmake-${CMAKE_VERSION}-linux-x86_64.sh -o cmake.sh \
+RUN dpkg --add-architecture arm64 && dpkg --add-architecture armhf
```
scareything, line 37: is this why you switched to Debian?

.github/actions/openziti-tunnel-build-action/ubuntu-20.04/Dockerfile
qrkourier: It allows us to use Debian MultiArch which is not available in Ubuntu (binary deps not available for foreign archs). This leads to a simpler cross-compile environment setup that's easy to express in a Dockerfile while remaining compatible with Ubuntu.

.github/actions/openziti-tunnel-build-action/ubuntu-20.04/Dockerfile
qrkourier: Yes. You can add foreign arch in Ubuntu, but `apt update` breaks immediately, and adding a new repo sources list for Ubuntu ports doesn't unbreak `apt update`.

scareything (approved):

PR #547, opened 2022-11-07
Fix bug where log doesn't rotate at the end of a month.

### LINKED ISSUE #486: log doesn't rotate on the end of month
log_utils.c has a problem on end of month. it uses tm_mday to determine if a log roll should occur and at the end of the month the day will be 28/29/30/31 of the 'orig_time' and the next time will be 1 since its' the "day in the month". The logs are still there, but the log file doesn't rotate.

Fix that. See line 112 ish:
```
        if (orig_time->tm_mday < tm->tm_mday) {
```

szymonnogiec:
Resolves #486. Check also a month and year difference (will prevent a bug where log would not be rotated at the end of the year).

### DIFF
--- a/programs/ziti-edge-tunnel/windows/log_utils.c
+++ b/programs/ziti-edge-tunnel/windows/log_utils.c
@@ -109,7 +109,7 @@ void flush_log(uv_check_t *handle) {
 
     if (handle->data) {
         struct tm *orig_time = handle->data;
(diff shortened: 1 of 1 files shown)


szymonnogiec:
I'm on Linux, so I could not test that, but I assume it should be ok.

scareything (approved):

PR #580, opened 2023-01-06
describe the differences between the two containers

qrkourier:


scareything (approved):

qrkourier:
I'll merge this as is in response to the Discourse post about needing a summary of the two containers near the top. I'm also working on deleting some things to further simplify the container build and will send another PR for docs changes with that.

PR #588, opened 2023-01-12
fix tunneler container readme links for Docusaurus import

qrkourier:
so that we can import the Markdown into Docusaurus


docker/README.md
```diff
@@ -27,7 +27,7 @@ You may bind a host directory to the container filesystem in `/ziti-edge-tunnel`
 
 ## Use Case: Hosting OpenZiti Services
 
-This use case involves deploying the OpenZiti tunneler as a reverse proxy to publish regular network servers to your OpenZiti Network. You may locate the published servers in a Docker bridge network (use network mode `bridge`) or the Docker host's network (use network mode `host`). See [the Linux tunneler doc](https://openziti.github.io/ziti/clients/linux.html) for general info about the OpenZiti tunneler. Use the `openziti/ziti-host` container image for this case.
+This use case involves deploying the OpenZiti tunneler as a reverse proxy to publish regular network servers to your OpenZiti Network. You may locate the published servers in a Docker bridge network (use network mode `bridge`) or the Docker host's network (use network mode `host`). See [the Linux tunneler doc](https://openziti.github.io/search/?q=linux+tunnel) for general info about the OpenZiti tunneler. Use the `openziti/ziti-host` container image for this case.
```
dovholuknf, line 30: https://openziti.github.io/search/?q=linux+tunnel ? this should take you to the page - not to search

docker/README.md
```diff
@@ -68,7 +68,7 @@ docker run \
   openziti/ziti-host
 ```
 
-This example uses [the included Docker Compose project](./docker-compose.yml) to illustrate publishing a server container to your OpenZiti Network.
+This example uses [the included Docker Compose project](https://github.com/openziti/ziti-tunnel-sdk-c/blob/main/docker/docker-compose.yml) to illustrate publishing a server container to your OpenZiti Network.
```
dovholuknf, line 71: i don't like these changes. using the full url in this way doesn't feel worth it

qrkourier: It's essential for including this README in the Docusaurus site because the build doesn't know about relative links.

qrkourier: @dovholuknf and I found that Docusaurus will in fact render relative links in some situations. For example, this worked:

```md
[link to Dockerfile](Dockerfile.ziti-host)
```

While this did not work:

```md
[link to Dockerfile](./Dockerfile.ziti-host)
```

docker/README.md
```diff
@@ -118,7 +118,7 @@ This example uses [the included Docker Compose project](./docker-compose.yml) to
     ZITI_IDENTITY_JSON="$(< /tmp/my-ziti-id.json)" docker-compose run ziti-host run-host --verbose=4
     ```
 
-1. Access the demo server via your OpenZiti Network: [http://hello-docker.ziti](http://hello-docker.ziti)
+1. Access the demo server via your OpenZiti Network: `http://hello-docker.ziti`
```
dovholuknf, line 121: this was clickable before if ziti was running and it's not now. i think it's better as a link?

dovholuknf (approved):
i personally like the link.

docker/README.md
qrkourier: It has potential value as a clickable link, but that requires the reader to know they need to have it running for the link to work, it will get detected as a broken link by the link checker bot when it's imported to the docs site.

PR #593, opened 2023-01-27
flush dns cache on Windows when starting and stopping

### LINKED ISSUE #590: flush DNS on start, maybe on stop
When a tunneler starts, it needs to flush dns to ensure that users get access to their services when the user expects to be able to access them (when the UI/status shows running). To do that, we need to flush dns on startup. This used to be part of ZDEW but was lost in the move to the c-based code.

Add flush to the startup of linux, windows and mac-based code.

Consider adding flush to the 'stop' event as well. It seems like a friendly thing to do when stopping the client as well

scareything:
fixes #590

### DIFF
--- a/programs/ziti-edge-tunnel/netif_driver/linux/tun.c
+++ b/programs/ziti-edge-tunnel/netif_driver/linux/tun.c
@@ -206,7 +206,7 @@ static void dns_update_systemd_resolve(const char* tun, unsigned int ifindex, co
     int s = run_command_ex(false, "systemd-resolve --status | fgrep  'DNS Domain' | fgrep -q '~.'");
     // set wildcard domain if any other resolvers set it.
     if (s == 0) {
-        run_command(SYSTEMD_RESOLVE " -i %s --set-doamin='~.'", dns_maintainer.tun_name);
+        run_command(SYSTEMD_RESOLVE " -i %s --set-domain='~.'", dns_maintainer.tun_name);
     } else {
         // Use busctl due to systemd version differences fixed in systemd>=240
         run_command(BUSCTL " call %s %s %s SetLinkDomains 'ia(sb)' %u 0",
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -1570,7 +1570,9 @@ static int run_tunnel(uv_loop_t *ziti_loop, uint32_t tun_ip, uint32_t dns_ip, co
         }
     }
     run_tunneler_loop(ziti_loop);
-    tun->close(tun->handle);
+    if (tun->close) {
+        tun->close(tun->handle);
+    }
     return 0;
 }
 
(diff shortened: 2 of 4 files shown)


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1618,9 +1617,6 @@ static void run_tunneler_loop(uv_loop_t* ziti_loop) {
     }
 
     free(tunneler);
-#if _WIN32
-    close_log();
```
scareything, line 1622: this was causing a crash because it nulls the logger file pointer, and logging happens after this point. without crashes, the remainder of `run_tunnel` (after `uv_run` returns) is executed.

programs/ziti-edge-tunnel/netif_driver/linux/tun.c
```diff
@@ -206,7 +206,7 @@ static void dns_update_systemd_resolve(const char* tun, unsigned int ifindex, co
     int s = run_command_ex(false, "systemd-resolve --status | fgrep  'DNS Domain' | fgrep -q '~.'");
     // set wildcard domain if any other resolvers set it.
     if (s == 0) {
-        run_command(SYSTEMD_RESOLVE " -i %s --set-doamin='~.'", dns_maintainer.tun_name);
+        run_command(SYSTEMD_RESOLVE " -i %s --set-domain='~.'", dns_maintainer.tun_name);
```
dovholuknf, line 209: i feel like this must fix 'some other bug'?

dovholuknf (approved):

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1573,6 +1571,7 @@ static int run_tunnel(uv_loop_t *ziti_loop, uint32_t tun_ip, uint32_t dns_ip, co
         }
     }
     run_tunneler_loop(ziti_loop);
+    tun->close(tun->handle);
```
scareything, line 1574: @smilindave26 the tsdk is now closing the netif_driver when the uv_loop is stopped. I think the apple tunneler will be OK with this, but wanted to point it out.

programs/ziti-edge-tunnel/netif_driver/linux/tun.c
scareything: yes. without this change, our dns server will not be consulted when all of the following conditions are met:

- on Linux
- when the dns maintainer detects that `systemd-resolve` should be used to update the system's resolver
- and the global resolver has the "wildcard" dns route

with that said, this is an unlikely set of circumstances. the dns maintainer looks for update methods in this order:

1. sd-bus direct API calls
2. /usr/bin/resolvectl
3. /usr/bin/systemd-resolve

so this will only make a difference when the api and resolvectl methods fail. I think we've written the exact versions up somewhere (maybe @qrkourier or @sabedevops can comment), but on any reasonably recent linux distro the sd-bus API will be selected.

dovholuknf: Sorry. My only point is that it's a different bug and not pertaining to windows is all.

PR #595, opened 2023-02-01
Use ziti conn bridge for hosting

ekoby:


lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -35,10 +35,19 @@
...
-#define MAX_OUTSTANDING_WRITES 8
-
 /********** hosting **********/
+static void on_bridge_close(uv_handle_t *handle);
+
+struct hosted_io_ctx_s {
+    struct hosted_service_ctx_s *service;
+    ziti_connection client;
+    char server_dial_str[64];
+    int server_proto_id;
```
scareything, line 45: Do we actually need `server_proto_id` any more?

scareything (approved):

lib/ziti-tunnel-cbs/ziti_hosting.c
ekoby: cleaned up

PR #600, opened 2023-02-13
set user after privileged copy command

qrkourier:
and optimize bash entrypoint


qrkourier:
@scareything May I please have a review?

dovholuknf (approved):

PR #607, opened 2023-02-27
PERF: Implement TCP back-pressure

ekoby:
- limit the amount of outstanding written data
- linux: read more than one packet on each loop iteration -- uv_poll_t is triggered once per loop iteration


lib/ziti-tunnel-cbs/ziti_tunnel_cbs.c
```diff
@@ -292,6 +294,9 @@ void * ziti_sdk_c_dial(const void *intercept_ctx, struct io_ctx_s *io) {
         return NULL;
     }
     io->ziti_io = ziti_io_ctx;
+    ziti_io_ctx->ziti_eof = false;
```
scareything, line 297: maybe it's time to use `calloc`?

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
```diff
@@ -262,11 +262,14 @@ DECLARE_MODEL(api_event, ZTX_API_EVENT_MODEL)
...
 typedef struct tunneled_service_s tunneled_service_t;
 
+#define MAX_WRITE_SIZE (128 * 1024)
+
 /** context passed through the tunneler SDK for network i/o */
 typedef struct ziti_io_ctx_s {
     ziti_connection      ziti_conn;
     bool ziti_eof;
     bool tnlr_eof;
+    uint64_t active_writes;
```
scareything, line 272: it looks like this is actually being used to track pending bytes? if so, this name is misleading.

scareything (approved):
I'll take a look at the windows build

lib/ziti-tunnel-cbs/include/ziti/ziti_tunnel_cbs.h
ekoby: ahh, much better name

scareything (approved):

PR #614, opened 2023-03-10
Run with reduced privileges when started by systemd

scareything:
This is essentially @tomc797's [recent PR](#599), slightly massaged so that the socket files don't require root/wheel access (if a 'ziti' group exists), and also support the older redhat distros we still build on.

cc @sabedevops, @qrkourier


scareything:
release notes content

# changes

The rpm and debian packages now create a non-administrative 'ziti' user and group, and the systemd unit for the
`ziti-edge-tunnel` service uses this user to run the ziti-edge-tunnel process. The process is granted `CAP_NET_ADMIN`
capability so that it can create the tun device and manipulate routes. PolicyKit is also configured to permit the 'ziti'
user to configure link DNS.

# breaking changes

The location of the IPC sockets that ziti-edge-tunnel creates has changed from "/tmp" to "/tmp/.ziti" when running on
Linux and Darwin. Furthermore, the /tmp/.ziti directory is only accessible by owner and group (mode 0x750), so processes
that use the IPC sockets must be started by users that are members of the /tmp/.ziti directory's group.

When `ziti-edge-tunnel` detects that it is running as root (e.g. via sudo, and/or otherwise _not_ via systemd), it looks
for a local 'ziti' group. If found, the group of /tmp/.ziti is changed to 'ziti', so that users who are members of the
'ziti' group may use the IPC sockets.

You can create a 'ziti' user and group if you're starting `ziti-edge-tunnel` as root. and you wish to use the IPC
sockets from processes that are not owned by root:

## Linux

    # useradd --system --home-dir=/var/lib/ziti --comment 'openziti user' --user-group ziti

## Darwin

    # dseditgroup -o create ziti
    # sysadminctl -addUser ziti -shell /usr/bin/false -GID $(dscl . -read /groups/ziti PrimaryGroupID | awk '{print $2}')

qrkourier:
I can live with those breaking changes for sure! Happy to see this. I would add `--shell /usr/sbin/nologin` for Linux. I would add a Linux hint for adding yourself to the group because it's easy to make the mistake of forgetting `usermod --append` and unintentionally revoking all preexisting group grants.

```bash
usermod --append --groups ziti $USER
```

qrkourier:
Looks like `/usr/sbin/login` is Debian-centric, so `/sbin/login` is more portable, where's `/bin/false` is the most portable. Both `nologin` binaries are identical on my Debian-based system.

```bash
❯ for TAG in \
  amazonlinux:{1,2} oraclelinux:{7..9} debian:{buster,bullseye,bookworm} \
  registry.access.redhat.com/ubi{8,9}/ubi ubuntu:{bionic,focal,jammy} \
  quay.io/centos/centos:7 fedora:{34,35,36} rockylinux/rockylinux:{8,9}; do
    docker run --rm --platform amd64 $TAG \
      bash -c '
        if which nologin &>/dev/null; then
          echo '"$TAG: has "'$(which nologin);
        fi
        for BIN in /bin/false {/usr,}/sbin/nologin; do
            if [[ -x $BIN ]]; then
              echo '"$TAG: has "'$BIN;
            else
              echo '"$TAG: missing "'$BIN;
            fi
        done';
      done | sort -u
amazonlinux:1: has /bin/false
amazonlinux:1: missing /sbin/nologin
amazonlinux:1: missing /usr/sbin/nologin
amazonlinux:2: has /bin/false
amazonlinux:2: missing /sbin/nologin
amazonlinux:2: missing /usr/sbin/nologin
debian:bookworm: has /bin/false
debian:bookworm: has /sbin/nologin
debian:bookworm: has /usr/sbin/nologin
debian:bullseye: has /bin/false
debian:bullseye: has /usr/sbin/nologin
debian:bullseye: missing /sbin/nologin
debian:buster: has /bin/false
debian:buster: has /usr/sbin/nologin
debian:buster: missing /sbin/nologin
fedora:34: has /bin/false
fedora:34: has /sbin/nologin
fedora:34: has /usr/sbin/nologin
fedora:35: has /bin/false
fedora:35: missing /sbin/nologin
fedora:35: missing /usr/sbin/nologin
fedora:36: has /bin/false
fedora:36: missing /sbin/nologin
fedora:36: missing /usr/sbin/nologin
oraclelinux:7: has /bin/false
oraclelinux:7: has /sbin/nologin
oraclelinux:7: has /usr/sbin/nologin
oraclelinux:8: has /bin/false
oraclelinux:8: has /sbin/nologin
oraclelinux:8: has /usr/sbin/nologin
oraclelinux:9: has /bin/false
oraclelinux:9: has /sbin/nologin
oraclelinux:9: has /usr/sbin/nologin
quay.io/centos/centos:7: has /bin/false
quay.io/centos/centos:7: has /sbin/nologin
quay.io/centos/centos:7: has /usr/sbin/nologin
registry.access.redhat.com/ubi8/ubi: has /bin/false
registry.access.redhat.com/ubi8/ubi: has /sbin/nologin
registry.access.redhat.com/ubi8/ubi: has /usr/sbin/nologin
registry.access.redhat.com/ubi9/ubi: has /bin/false
registry.access.redhat.com/ubi9/ubi: has /sbin/nologin
registry.access.redhat.com/ubi9/ubi: has /usr/sbin/nologin
rockylinux/rockylinux:8: has /bin/false
rockylinux/rockylinux:8: has /sbin/nologin
rockylinux/rockylinux:8: has /usr/sbin/nologin
rockylinux/rockylinux:9: has /bin/false
rockylinux/rockylinux:9: has /sbin/nologin
rockylinux/rockylinux:9: has /usr/sbin/nologin
ubuntu:bionic: has /bin/false
ubuntu:bionic: has /usr/sbin/nologin
ubuntu:bionic: missing /sbin/nologin
ubuntu:focal: has /bin/false
ubuntu:focal: has /sbin/nologin
ubuntu:focal: has /usr/sbin/nologin
ubuntu:jammy: has /bin/false
ubuntu:jammy: has /sbin/nologin
ubuntu:jammy: has /usr/sbin/nologin
```

sabedevops:
I'd consider systemd Dynamic Users for any of the newer distros that have `systemd` version >= 232: 

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1597,13 +1597,52 @@ static int make_socket_path(uv_loop_t *loop) {
...
+           *   - warn that IPC sockets are not being started because admin privs would be needed to access the sockets.
+           *   - include helpful commands to create ziti user/group
+           */
+          if (geteuid() == 0 && req.statbuf.st_gid == getegid()) {
+              bool chgrp_succeeded = false;
+              ZITI_LOG(DEBUG, "created IPC socket directory %s with root ownership. attempting to set group to 'ziti'",
+                       SOCKET_PATH);
+              struct passwd *pwd = getpwnam("ziti");
+              if (pwd) {
+                  rc = chown(SOCKET_PATH, -1, pwd->pw_gid);
```
tomc797: HP-UX and other unixes don't check permissions on sockets. That's why the socket needs to be in a directory with the correct permissions.

tomc797:
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=: 

> Care should be taken that any processes running as part of a unit for which dynamic users/groups are enabled do not leave files or directories owned by these users/groups around, as a different unit might get the same UID/GID assigned later on, and thus gain access to these files or directories.

The tunneler creates files in ``/var/lib/ziti`` which would be abandoned if DynamicUser was created.

tomc797:
@scareything Instead of ``chown(ziti)`` of ``SOCKET_PATH``, would a more comprehensive approach be set the primary group to ``ziti``, that is, ``setgid(ziti)``?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1597,13 +1597,52 @@ static int make_socket_path(uv_loop_t *loop) {
...
+           * sudo; we also don't want to leave the IPC socket directory wide open.
+           *
+           * Look for a "ziti" user/group. If found:
+           *   - chgrp the directory to the ziti group
+           *   - log that the sockets can be used by users in the "ziti" group. include helpful `usermod` command
+           * If "ziti" group is not found:
+           *   - warn that IPC sockets are not being started because admin privs would be needed to access the sockets.
+           *   - include helpful commands to create ziti user/group
+           */
+          if (geteuid() == 0 && req.statbuf.st_gid == getegid()) {
```
tomc797, line 1619: Should the ``chown`` be attempted when not root? What about the case that the user has ``ziti`` as a supplementary group?

sabedevops:
> https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=:
> 
> > Care should be taken that any processes running as part of a unit for which dynamic users/groups are enabled do not leave files or directories owned by these users/groups around, as a different unit might get the same UID/GID assigned later on, and thus gain access to these files or directories.
> 
> The tunneler creates files in `/var/lib/ziti` which would be abandoned if DynamicUser was created.

Please see `StateDirectory=` here: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=. 

In this case, if `StateDirectory=ziti`, the process has access `/var/lib/ziti`. When `DynamicUser=true`, the UID is actually stably derived from the hash of the service name, and should not change. In cases where it does, the directory structure is automatically `chown`'d. 

 I should be clear that this option bumps the requirement to systemd >= 235. 
 
 Furthermore, `PrivateTmp=true` is implied with `DynamicUser=true`. Placing the runtime socket there is advantageous.

tomc797:
> > https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=:
> > > Care should be taken that any processes running as part of a unit for which dynamic users/groups are enabled do not leave files or directories owned by these users/groups around, as a different unit might get the same UID/GID assigned later on, and thus gain access to these files or directories.
> > 
> > 
> > The tunneler creates files in `/var/lib/ziti` which would be abandoned if DynamicUser was created.
> 
> Please see `StateDirectory=` here: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=.
> 
> In this case, if `StateDirectory=ziti`, the process has access `/var/lib/ziti`. When `DynamicUser=true`, the UID is actually stably derived from the hash of the service name, and should not change. In cases where it does, the directory structure is automatically `chown`'d.
> 
> I should be clear that this option bumps the requirement to systemd >= 235.
> 
> Furthermore, `PrivateTmp=true` is implied with `DynamicUser=true`. Placing the runtime socket there is advantageous.

Neat feature. The only limitation is that PolicyKit looks up users when evaluating the policy. 

> > https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=:
> > > Care should be taken that any processes running as part of a unit for which dynamic users/groups are enabled do not leave files or directories owned by these users/groups around, as a different unit might get the same UID/GID assigned later on, and thus gain access to these files or directories.
> > 
> > 
> > The tunneler creates files in `/var/lib/ziti` which would be abandoned if DynamicUser was created.
> 
> Please see `StateDirectory=` here: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=.
> 
> In this case, if `StateDirectory=ziti`, the process has access `/var/lib/ziti`. When `DynamicUser=true`, the UID is actually stably derived from the hash of the service name, and should not change. In cases where it does, the directory structure is automatically `chown`'d.
> 
> I should be clear that this option bumps the requirement to systemd >= 235.
> 
> Furthermore, `PrivateTmp=true` is implied with `DynamicUser=true`. Placing the runtime socket there is advantageous.

Neat feature. I wonder how this interacts with ``PolicyKit`` rules. I would assume ``PolicyKit`` uses ``/etc/passwd`` and related files to identify users.

sabedevops:
@tomc797 It ultimately relies on `nsswitch` and the configured `passwd` and `group` databases. Thinking about it this way, I suspect `DynamicUser` would only work in this case if the `libnss-systemd` modules for each of these databases are enabled. That's a no-go, I think. Thanks for pointing out that deficiency.

The more I think about it, the more I wonder if we should consider the ZET a system service, given its inherently privileged nature. In that case, `system-sysusers` might be the right facilitator. It allows us to declaratively drop in a conf file to like `/usr/lib/sysusers.d/ziti.conf` and directly manipulates the `/etc/passwd` and `/etc/group` database files directly. This is the approach taken by `systemd-resolved`, and `httpd`, and should avoid the issue you brought up. This also reduces the requirement to systemd >=215, which I believe satisfies all currently supported distros.

Excited about this PR, thanks for your hard work. 

P.S. This reply is partially written as documentation for the future, in case it doesn't make it to this PR.

dovholuknf (approved):

qrkourier:
Congratulations on merging this! Glad to see it.

PR #620, opened 2023-03-22
build with OpenSSL instead of Mbet-TLS

qrkourier:


qrkourier:
This changes the build in the following ways:
1. start building the arm, arm64, and amd64 executable binary artifacts with OpenSSL instead of Mbed-TLS
2. start publishing an arm64 DEB for Ubuntu Bionic

The new DEB is in this branch because the solution for cross-compiling the executable binary artifact is the same as building the DEB package for Bionic on arm64.

ekoby (approved):

PR #628, opened 2023-04-11
Use uv_getaddrinfo when resolving hostnames in host configurations

### LINKED ISSUE #626: hostnames in host.v1 configuration are sometimes not resolved
A user encountered hostname resolution failures for their hosted server address, even though the address could be resolved by other processes on the system (e.g. ping, etc.). The tunneler was running on Windows, probably a server edition, and logged the following error:

    ERROR tunnel-cbs:ziti_hosting.c:394 on_hosted_client_connect() hosted_service[vault.service], client[Laptop]: getaddrinfo(myapp.com,443) failed: No such host is known. 

`nslookup` and `ping` were able to resolve "myapp.com".

discourse thread: https://openziti.discourse.group/t/ziti-tunneler-on-windows-not-working/1196

scareything:
Recently I took a closer look at the hosting code and realized that we're using `getaddrinfo` to resolve hostnames. `libuv` provides this function so we should use it.

There's a slight chance that this would fix #626, but no promises.


dovholuknf (approved):
looks ok to me

Issue #569, opened 2022-12-05
How to configure log timestamp?


qrkourier:
`ziti-edge-tunnel run` and `run-host` emit log messages prefixed with a timestamp expressed as `ss.ms` elapsed since the process started. It would be convenient for comprehending logs if the timestamp conformed to a readable datetime standard like IS8601, or perhaps another standard that expresses milliseconds. Is there a way to influence the format of this field? I did try setting `PFXLOG_NO_JSON="true"` just in case that would have an effect.

dovholuknf:
you're asking for when you're not using journald i assume, and when you're just running it directly? `PFXLOG_NO_JSON` is not relevant to the ziti-edge-tunnel

qrkourier:
Correct, and I see what you mean. `systemd-journald` interprets the `ss.ms` output and prints a readable datetime like `Dec 05 12:59:24`. My `run-host` deployments aren't managed by `systemd`, and so there's no interpretation of the `ss.ms` value in messages to stdout.

dariuszSki:
@ekoby @scareything can we implement this? I am troubleshooting a problem and this option would be useful when  a user redirect stdout/err to a file.  This is the case now with this user. It is difficult to correlate the errors generated by ZET to errors seen by the rest of the networks like ERs/Controller.

ekoby:
It's already available. Set environment variable `ZITI_TIME_FORMAT=utc`

PR #635, opened 2023-04-24
add a script to parse the ms offset in log messages

qrkourier:
workaround for https://github.com/openziti/ziti-tunnel-sdk-c/issues/569

Run it like this:

```bash
./scripts/timestamp.sh /tmp/ziti-edge-tunnel.log
```

### DIFF
--- a/scripts/timestamp.sh
+++ b/scripts/timestamp.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+(( ${#} )) && {
+    LOGFILE="${1}"
+    shift
+}
+
+(( ${#} )) && {
+    echo "ERROR: too many arguments" >&2
+    exit 1
+}
+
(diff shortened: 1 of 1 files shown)


ekoby:
this is not needed. zitisdk logging can be configured to put UTC timestamps
set `ZITI_TIME_FORMAT=utc`

qrkourier:
No one likes this ugly, awful script that solves the wrong problem, which I completely understand. Still, it's super useful to me and to support for parsing a user-provided log file that doesn't have timestamps. I think we should commit it for that reason alone. I'll keep a copy handy and share it whenever this problem creeps up again in case no one approves this PR. This was in response to a real-world need, not just ignorance about configuring log output.

scripts/timestamp.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
scareything, line 1: How about a comment here to explain what this script does?

scareything (approved):

PR #647, opened 2023-05-05
Adjusts systemd unit directory for older Ubuntu distribution packages.

sabedevops:


programs/ziti-edge-tunnel/package/CPackPackage.cmake
```diff
@@ -60,7 +60,9 @@ set(SYSTEMD_EXECSTARTPRE "${SYSTEMD_SERVICE_NAME}.sh")
 set(SYSTEMD_ENV_FILE "${SYSTEMD_SERVICE_NAME}.env")
 
 set(SYSTEMD_SYSTEM_ETC "/etc/systemd/system")
-set(SYSTEMD_UNIT_DIR "/usr/lib/systemd/system")
```
scareything, line 63: could /lib/systemd work on all distros? (it would work because of the /lib -> /usr/lib symlink on more recent distros)

scareything (approved):

programs/ziti-edge-tunnel/package/CPackPackage.cmake
sabedevops: It's probably the better approach, actually. I can update the PR. `dh_installsystemd` actually does install to `/lib`. The rpm macros expand to `/usr/lib`, but I'm not sure there is any advantage to the added complexity.

PR #650, opened 2023-05-06
Allocate tun network device with name "ziti0"

tomc797:
Allocate a tun network device with name ``ziti0``. 

When allocating a device, ``ifr.ifr_name`` accepts a format string with the ``%d`` specifier. The OS then will allocate a device with the generated name. The field ``ifr.ifr_name`` is overwritten with name and returned to the user.


scareything (approved):
Thanks again for your contributions! I knew that we could set the tun device name but I didn't know you could use a format spec to let the name include the interface index. Nice trick!

PR #662, opened 2023-05-17
build for Linux in ziti-builder container

qrkourier:


qrkourier:
@ekoby Ready for review.

.github/workflows/cmake.yml
```diff
@@ -1,128 +1,77 @@
...
             name: Linux x86_64
-            install: $common_ubuntu_deps build-essential
-            # build current openssl libs instead of using old versions provided by apt on ubuntu 18
-            cmake_opts: "`-DVCPKG_OVERLAY_PORTS=./vcpkg-overlays/linux-syslibs/ubuntu18`"
-            target: linux-x64
+            preset: linux-x64
 
           - os: ubuntu-20.04
-            container: ubuntu:18.04
+            container: openziti/ziti-cmake
```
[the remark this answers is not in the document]
qrkourier: Thanks for catching this. I've corrected the image reference to `openziti/ziti-builder`. `ziti-cmake` was the old name.

ziti-builder.sh
```diff
@@ -0,0 +1,130 @@
...
+    else
+        [[ -d ./build ]] && rm -rf ./build
+        cmake \
+            -E make_directory \
+            ./build  
+        cmake \
+            --preset "${CMAKE_PRESET:-ci-linux-x64}" \
+            -DCMAKE_BUILD_TYPE="${CMAKE_CONFIG:-Release}" \
+            -DBUILD_DIST_PACKAGES="${BUILD_DIST_PACKAGES:-OFF}" \
+            -DVCPKG_OVERLAY_PORTS="./vcpkg-overlays/linux-syslibs/ubuntu18" \
```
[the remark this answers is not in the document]
qrkourier: This script always runs on Ubuntu 18.04 because that's the source image for the builder image `ziti-builder`, and this script always builds the "bundle" target, not "package," when running in CI because we have the distro-specific builder images for target "package."

I was thinking that a dev might want to build target "package" locally, so I left that option in the script. Does it answer your concern to know that package artifacts built with this script are never released? If so, should we make the script more flexible to use the default preset for local builds of target "package" or "bundle" or both?

.github/workflows/cmake.yml
```diff
@@ -1,128 +1,77 @@
...
-
-      - uses: lukka/run-vcpkg@v10
-        with:
-          # use 2023.02.24 vcpkg baseline,
-          # see https://learn.microsoft.com/en-us/vcpkg/users/examples/versioning.getting-started#builtin-baseline
-          vcpkgGitCommitId: 'a7b6122f6b6504d16d96117336a0562693579933'
+      - name: Run CMake for Linux
+        if: ${{ matrix.container != null }}
+        run: |
+          ./ziti-builder.sh \
```
ekoby, line 65: by using the script you're not taking advantage of vcpkg caching which reduces the build times considerably
also, it is not clear what VCPKG baseline/commit the script would be using

qrkourier: Re: binary caching for vcpkg

Could you let me know if you meant to suggest a persistent cache for GH Actions runs? If so, is that working for the zitify or ziti-sdk-c projects? I understand the hosted job runners start with a pristine filesystem, so it's necessary to self-host the runner to make other storage available. That's doable but isn't expedient unless build minutes become a scarce resource. I've used the artifact upload/download Actions to move files around, but not a cache directory.

You might have meant that a developer running the script locally doesn't benefit from their VCPKG binary cache. That's something I hadn't considered. If the cache is in a predictable file path, we could make it available inside the builder container easily. Is that what you had in mind?

---

Re: VCPKG baseline/commit

Will you please clarify this one? The VCPKG preset file is from the same commit as the script because they're part of the repo. I suspect you meant something else.

ekoby: GH workflows come with a way of caching _stuff_ between runs. Usually actions maintain their own caches.
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

For example `run-vcpkg` will restore 3rd party binaries from previous runs so that the current run does not have rebuild everything 

when you build locally, your VCPKG_ROOT serves as the cache (even if you clean your build directory 3rd party dependencies are repopulated from VCPKG_ROOT/packages)

ekoby: VCPKG baseline refers to git commit id for vcpkg repo. Essentially if fixes the versions of third party dependencies that we use.

ekoby: here is an example of cache reuse in ziti-sdk-c https://github.com/openziti/ziti-sdk-c/actions/runs/5270170584/jobs/9529313420#step:5:83

qrkourier: > it is not clear what VCPKG baseline/commit the script would be using

@ekoby If I understand correctly, the baseline you're referring to is defined in the `ziti-builder` container image [on this line](https://github.com/openziti/ziti-builder/blob/main/docker-image/Dockerfile#L69). Is it sufficient to pin the VCPKG baseline in this way, or do you see a need to configure the VCPK version in the workflow runs that consume the `ziti-builder` image?

.github/workflows/cmake.yml
```diff
@@ -1,128 +1,77 @@
...
+            -p "ci-${{ matrix.preset }}" \
+            -t bundle
 
-      - uses: lukka/run-cmake@v10
-        name: Configure CMake
+      - name: Run CMake for non-Linux platforms
```
ekoby, line 69: I don't like that we have two different build flows -- Linux(script) vs non-Linux(action) -- they can get out of sync if steps are added to the build

ekoby (dismissed):

.github/workflows/cmake.yml
qrkourier: The uniformity of Linux CI and local Linux builds seems to be less important than I thought believed when I proposed `ziti-builder.sh`. Now there's more favor for uniformity of Linux and non-Linux in CI, so we should let this T-SDK build reflect the approach used in the C-SDK, a composite Action used by both Linux and non-Linux builds, and relocate the script in `./scripts/ziti-builder.sh` so that it can serve whomever who wants to build locally with the builder container, myself included. This will also solve caching and VCPKG baseline in the same way as C-SDK.

Issue #474, opened 2022-08-09
specific routes should not be created for IPs in the dns range


scareything:
The tSDK adds a subnet route (/10 by default) to the tun interface to handle IPs that are given to services with configs that have hostnames in their address list.

`ziti_tunneler_intercept` / `add_route` currently adds superfluous "/32" routes for any IPs that were handed out by the DNS server.

PR #676, opened 2023-06-08
Disregard route if update is a subnet of tun's address

tomc-xsec:
As the ```tun_ip``` is derived from ```dns_block```, set the ```tun```'s address prefix length equal to ```dns_block``` prefix length. Then ignore any route updates that are a subnet of the ```tun```'s address. This becomes helpful when large number of DNS intercepts are employed (say wildcard), each causing a /32 route to be added.

### DIFF
--- a/programs/ziti-edge-tunnel/netif_driver/linux/tun.c
+++ b/programs/ziti-edge-tunnel/netif_driver/linux/tun.c
@@ -126,6 +126,36 @@ static void route_updates_done(uv_work_t *wr, int status) {
     free(wr);
 }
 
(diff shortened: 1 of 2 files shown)


scareything:
Thanks for taking a stab at this. This has been on our minds for a while now (https://github.com/openziti/ziti-tunnel-sdk-c/issues/474).

Ideally the matching routes would be caught at the tunneler-sdk layer (probably in `add_route`) so all platforms benefit. Da you have any interest in doing the route detection there? I'd think `ziti_address_match` would be helpful in the tsdk.

PR #683, opened 2023-06-29
Feature/remove sed runtime dependency

sabedevops:


programs/ziti-edge-tunnel/netif_driver/linux/resolvers.c
```diff
@@ -468,49 +502,74 @@ void dns_update_etc_resolv(const char* tun, unsigned int ifindex, const char* ad
...
+          if (fread(remaining_content, sizeof(char), (size_t)remaining_size, file) != remaining_size) {
               if (ferror(file) || feof(file)) {
                   ZITI_LOG(ERROR, "Error during file stream operation or EOF received.");
               }
+              cleanup_bufferp(&remaining_content);
+              CLEANUP_ETC_RESOLV();
+          }
+          remaining_content[remaining_size] = '\0';
+
+          _cleanup_(cleanup_bufferp) char *rptr = realloc(replace, (size_t)(replace_size + remaining_size + 1));
```
scareything, line 533: why use a new variable here? I thought the common `realloc` pattern was `p = realloc(p, size);` ?

scareything (approved):

programs/ziti-edge-tunnel/netif_driver/linux/resolvers.c
```diff
@@ -432,24 +457,33 @@ static void cleanup_bufferp(char **buffer) {
...
 
-void dns_update_etc_resolv(const char* tun, unsigned int ifindex, const char* addr) {
-      const char* match = "nameserver ";
-      int replace_size = strlen(match) + strlen(addr) + 1;
-      _cleanup_(cleanup_bufferp) char* replace = (char *)malloc(replace_size);
+void dns_update_etc_resolv(const char *tun, unsigned int ifindex, const char *addr) {
+      bool copy_r = make_copy(RESOLV_CONF_FILE, RESOLV_CONF_FILE ".bkp");
+
+      const char *match = "nameserver ";
+      off_t replace_size = strlen(match) + strlen(addr) + sizeof(char);
```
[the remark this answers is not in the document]
sabedevops: This is handled in #L466. Accounting for the null byte is always handled in the allocation functions in this code so that I could keep it consistent across `malloc` and `realloc` when computing the `realloc` size in #L533. In other words, `*_size` here indicates the size _without_ the null bytes.

programs/ziti-edge-tunnel/netif_driver/linux/resolvers.c
sabedevops: I had some doubts about correct usage here after careful reading of the man page (it has very quirky behavior), so I took inspiration from information shared here:

https://stackoverflow.com/questions/44789295/correct-use-of-realloc

Additionally, there is interplay with the `cleanup()` function guard here I'm being careful to consider so we don't double `free()` in the case where the allocated memory block is moved from it's original location since `realloc()` will `free()` it internally. See #L534-544

PR #686, opened 2023-07-05
added AllowedSourceAddress info to events channel output

r-caamano:


programs/ziti-edge-tunnel/include/model/dtos.h
```diff
@@ -82,6 +86,7 @@ XX(Id, string, none, Id, __VA_ARGS__) \
 XX(Name, string, none, Name, __VA_ARGS__) \
 XX(Protocols, string, array, Protocols, __VA_ARGS__) \
 XX(Addresses, tunnel_address, array, Addresses, __VA_ARGS__) \
+XX(AllowedSourceAddresses, allowed_source_address, array, AllowedSourceAddresses, __VA_ARGS__) \
```
ekoby, line 89: make it a `list` -- it is easier to deal with

ekoby: why not use `ziti_address` instead of a new type?

r-caamano: Not clear to me how I would do that.  Seems all the items in the tunnel events are newly defined types in model/dtos.h which in some cases are populated with information from ziti_address i.e tunnel_address.  However they do not use it as a type directly.  Maybe I'm not interpreting your suggestion correctly.

r-caamano: Also seems all the other multi item entries in model/dtos.h associated with a TUNNEL_SERVICE are defined as arrays i.e.
XX(Protocols, string, array, Protocols, __VA_ARGS__) 
XX(Addresses, tunnel_address, array, Addresses, __VA_ARGS__) 
XX(Ports, tunnel_port_range, array, Ports, __VA_ARGS__)

r-caamano:
If you wanted to overload tunnel_address you could add a Boolean that identified the address entry as an allowed source.  That would require apps that currently use the tunnel_address (i.e. desktop-edge-ui) to reference that bool when determining whether or not to display the address.

programs/ziti-edge-tunnel/instance.c
```diff
@@ -286,6 +287,42 @@ tunnel_port_range *getTunnelPortRange(ziti_port_range *zpr){
...
+        parse_ziti_host_cfg_v1(&cfg_v1, cfg_json, strlen(cfg_json));
+        size_t n = 0;
+        int j = 0;
+        ziti_address_array allowed_src_addrs = cfg_v1.allowed_source_addresses;
+        for (int x = 0; allowed_src_addrs != NULL && allowed_src_addrs[x] != NULL; x++) {
+            n++;
+        }
+        allowed_src_addr_arr = calloc(n + 1, sizeof(tunnel_address *));
+        for (int i = 0; i < n; i++) {
+            if (allowed_src_addrs[i]->type != ziti_address_cidr) {
```
scareything, line 305: minor: if this test happened in the loop above where `n` is calculated, you could allocate precisely enough memory for `allowed_src_addr_arr`.

scareything (approved):

programs/ziti-edge-tunnel/instance.c
r-caamano: The reason I did not was because I would have to run the same test again in the second loop since I don't have a way to know which entries were discarded from the original array and did not want redundant code and assumed it should be rare that someone actually puts a hostname instead of a prefix since it would be in error.  If you prefer it I can add it.  let me know.

r-caamano:
Closing PR due to commits not being signed.  Will submit as new PR shortly

PR #690, opened 2023-07-12
fix `run --refresh ` parameter

### LINKED ISSUE #689: ziti-edge-tunnel run: --refresh has no effect
running `ziti-edge-tunnel run --refresh 60 -I...` still has `/service-updates` polling every 10 seconds

ekoby:
[fixes #689]


lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -103,6 +103,11 @@ const ziti_tunnel_ctrl* ziti_tunnel_init_cmd(uv_loop_t *loop, tunneler_context t
 
 IMPL_ENUM(mfa_status, MFA_STATUS)
 
+void ziti_set_refresh_interval(unsigned long seconds) {
+    unsigned long old = refresh_interval;
```
scareything, line 107: is there any reason to keep the old value?

ekoby: no, I was originally returning the old value, but it has no real use.
I'll clean up

PR #702, opened 2023-08-07
used versioned builder image

qrkourier:


ekoby (approved):
maybe a variable for version so it is on one place?

qrkourier:
> maybe a variable for version so it is on one place?

OK, I'll work on that separately.

PR #704, opened 2023-08-10
Feature/libacl etc resolver enablement

sabedevops:


.github/actions/openziti-tunnel-build-action/redhat-7/Dockerfile
```diff
@@ -39,7 +42,7 @@ ENV VCPKG_ROOT=/usr/local/vcpkg
 ENV VCPKG_FORCE_SYSTEM_BINARIES=yes
 
 RUN cd /usr/local \
-    && git clone --branch 2023.04.15 https://github.com/microsoft/vcpkg \
+    && git clone --branch 2023.06.20 https://github.com/microsoft/vcpkg \
```
scareything, line 45: we'll need a corresponding PR with this change (and possibly others) in ziti-builder to cover the GH release builds.

sabedevops: Can do. Which others can you think of?

sabedevops: https://github.com/openziti/ziti-builder/pull/23

scareything: I'm not sure which other changes will be needed there. I was mainly thinking about comparing the packages that were added here and make sure they are present in ziti-builder?

Once you have a candidate builder image you'll need to reference its version in this PR, yeah?

sabedevops: The packages are present in ziti-builder now from a previous PR. We need a newly published image tag that includes those changes and the above PR (the version bump) so we can update .github/workflows/cmake.yml references to point to it.

+cc @qrkourier

qrkourier: I just released `ziti-builder:1.0.6`.

sabedevops: Build seems to have run successfully. Thanks @qrkourier.

programs/ziti-edge-tunnel/CMakeLists.txt
```diff
@@ -34,13 +34,29 @@ add_executable(ziti-edge-tunnel ziti-edge-tunnel.c ${NETIF_DRIVER_SOURCE} ${ZITI
...
+    pkg_check_modules(LIBCAP IMPORTED_TARGET "libcap")
+
+    if (LIBACL_FOUND)
+        target_compile_definitions(ziti-edge-tunnel PRIVATE HAVE_LIBACL_PKG)
+        target_link_libraries(ziti-edge-tunnel PRIVATE PkgConfig::LIBACL)
+    endif()
+
+    if (LIBCAP_FOUND)
+        target_compile_definitions(ziti-edge-tunnel PRIVATE HAVE_LIBCAP_PKG)
+	    target_link_libraries(ziti-edge-tunnel PRIVATE PkgConfig::LIBCAP)
```
scareything, line 50: alignment

vcpkg.json
```diff
@@ -2,6 +2,10 @@
   "name": "ziti",
   "version-semver": "1.0.0",
   "dependencies": [
+    {
```
scareything, line 5: This will cause vcpkg to download and build libacl for all linux builds. Would it be better to use/depend on the distro-provided packages where available? If so, we have a pattern for using distro dependencies in vcpkg-overlays/linux-syslibs.

programs/ziti-edge-tunnel/package/CPackGenConfig.cmake
```diff
@@ -1,12 +1,12 @@
...
 
 if(CPACK_GENERATOR MATCHES "RPM")
-    set(CPACK_RPM_BUILDREQUIRES "cmake >= ${CMAKE_MINIMUM_REQUIRED_VERSION}, systemd-devel, gawk, gcc-c++ >= 4.9, python3, openssl-devel, zlib-devel")
+    set(CPACK_RPM_BUILDREQUIRES "cmake >= ${CMAKE_MINIMUM_REQUIRED_VERSION}, systemd-devel, gawk, gcc-c++ >= 4.9, python3, openssl-devel, zlib-devel, libacl-devel")
     if(CPACK_OS_RELEASE_NAME IN_LIST CPACK_RPM_DISTRIBUTIONS AND CPACK_OS_RELEASE_VERSION VERSION_GREATER "7")
         list(APPEND CPACK_RPM_BUILDREQUIRES "systemd-rpm-macros")
     endif()
     set(CPACK_RPM_PACKAGE_SOURCES OFF)
-    set(CPACK_RPM_PACKAGE_REQUIRES "iproute, gawk, systemd, libatomic, openssl-libs, zlib, polkit")
+    set(CPACK_RPM_PACKAGE_REQUIRES "iproute, gawk, systemd, libatomic, openssl-libs, zlib, polkit, libcap")
```
scareything, line 9: no runtime dependency on `libacl`?

sabedevops: The answer here is related to the question above. In the case of static compilation, there is no other runtime dependency.

```
ldd build/programs/ziti-edge-tunnel/Debug/ziti-edge-tunnel
        linux-vdso.so.1 (0x00007ffd01f5a000)
        libcap.so.2 => /lib64/libcap.so.2 (0x00007f4efe7dc000)
        libz.so.1 => /lib64/libz.so.1 (0x00007f4efe7c2000)
        libatomic.so.1 => /lib64/libatomic.so.1 (0x00007f4efe7b8000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f4efe6d7000)
        libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f4efe6c5000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f4efe4e5000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f4efe7fd000)
```

vcpkg.json
sabedevops: Do we have a way to specify to `vcpkg` that it should conditionally build a dependency instead?  This port made working with `libacl` nice since the pkg-config file `libacl.pc` results in a build where `libacl.a` is statically linked.

programs/ziti-edge-tunnel/CMakeLists.txt
sabedevops: Fixed.

programs/ziti-edge-tunnel/package/CPackGenConfig.cmake
scareything: I think it's generally preferable to link with shared libraries when they are available though, unless there's a reason not to (e.g. GH release builds). We manage currently this (.so's in package binaries, .a's in GH binaries) by using a special cmake preset (ci-linux-static-libssl, which should be renamed) for the GH release build.

If we're using the static libacl.a from vcpkg then I would not think we need `libacl-devel` in `CPACK_RPM_BUILDREQUIRES`?

sabedevops: The reason here was to minimize unknowns in library versioning and implementations, specifically, when it came to OpenWRT builds (at least for common architectures). Additionally, if I remember correctly, older versions of libacl didn't ship a pkg-config file, which meant that you couldn't leverage the pkg-config cmake module used here. The port addresses both those issues. Correct, I'll remove it. That was probably cruft from before the switch.

scareything: Ok. this is the main reason why there are distro-specific subdirectories in vcpkg-overlays/linux-syslibs.

.github/actions/openziti-tunnel-build-action/redhat-8/entrypoint.sh
```diff
@@ -39,16 +39,17 @@ done
...
-    cmake -E make_directory ./build  
+    cmake -E make_directory ./build
     # allow unset for scl_source scripts
     set +u
     source scl_source enable gcc-toolset-10 \
         && cmake \
             --preset "${cmake_preset}" \
             -DCMAKE_BUILD_TYPE="${cmake_config}" \
             -DBUILD_DIST_PACKAGES=ON \
+            -DVCPKG_OVERLAY_PORTS="./vcpkg-overlays/linux-syslibs/default;./vcpkg-overlays/linux-syslibs/redhat8" \
```
scareything, line 50: I'd rather not combine the overlay directories as I think it can lead to surprises.

The intent with the "default" overlay directory is that it contains all of the packages that can be provided by the system for modern distros (e.g. the "ideal" case), and the distro-specific overlay directories exist for platforms that can't use one or more of the packages from "default".

sabedevops: The alternative is to replicate the default in each of those directories. Simply specifying the single overlay directory overrides the default for the preset altogether. This is what led me to wonder if our current implementation is working as intended for other places where this is specified.

PR #718, opened 2023-08-23
use openziti/hello-world container image

qrkourier:


qrkourier:
@dovholuknf This one doesn't have side effects, so we could go ahead and review it.

qrkourier:
@dovholuknf This is ready for review.

scareything (approved):

PR #723, opened 2023-08-29
document verbose container args

qrkourier:


docker/README.md
```diff
@@ -220,6 +234,23 @@ docker run \
     openziti/ziti-edge-tunnel
 ```
 
+#### Troubleshooting `openziti/ziti-edge-tunnel`
+
+You may pass additional args by supplying the `run` mode followed by args when the container is run.
+
+```bash
+docker run \
+    --name ziti-tun \
```
scareything, line 243: This name evokes `tun` devices for me, which feels incongruent with a container name. fwiw I'd prefer `ziti-edge-tunnel`

scareything (approved):

docker/README.md
qrkourier, line 243:
```suggestion
    --name ziti-edge-tunnel \
```

PR #727, opened 2023-09-12
install the service unit file instead of symlinking

### LINKED ISSUE #726: Enabling package on ubuntu 18.04 does not work correctly
When trying to enable the ziti-edge-tunnel package on Ubuntu 18.04 by issuing "sudo systemctl enable ziti-edge-tunnel.service --now I get the following error:

The unit files have no installation config (WantedBy, RequiredBy, Also, Alias
settings in the [Install] section, and DefaultInstance for template units).
This means they are not meant to be enabled using systemctl.
Possible reasons for having this kind of units are:
1) A unit may be statically enabled by being symlinked from another unit's
   .wants/ or .requires/ directory.
2) A unit's purpose may be to act as a helper for some other unit which has
   a requirement dependency on it.
3) A unit may be started when needed via activation (socket, path, timer,
   D-Bus, udev, scripted systemctl call, ...).
4) In case of template units, the unit is meant to be enabled with some
   instance name specified.

If I copy /lib/systemd/system/ziti-edge-tunnel.service to /etc/systemd/system/ I am then able to successfully enable the service.

scareything:
fixes #726


qrkourier (approved):
It looks like this will fix Ubuntu Bionic's service unit.

PR #745, opened 2023-10-11
set required options in init_ziti_instance

scareything:


lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -677,15 +672,44 @@ struct ziti_instance_s *new_ziti_instance(const char *identifier) {
 }
 
 int init_ziti_instance(struct ziti_instance_s *inst, const ziti_config *cfg, const ziti_options *opts) {
+    FREE(inst->ztx);
```
dovholuknf, line 675: this seems like a strange, "first thing" to do, but i assume it's on purpose? just want to make sure it is?

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -677,15 +672,44 @@ struct ziti_instance_s *new_ziti_instance(const char *identifier) {
...
     rc = ziti_context_set_options(inst->ztx, opts);
+    if (rc != ZITI_OK) {
+        ZITI_LOG(ERROR, "ziti_context_set_options failed: %s", ziti_errorstr(rc));
+        FREE(inst->ztx);
+        return rc;
+    }
+
+    rc = set_tnlr_options(inst);
+    if (rc != ZITI_OK) {
+        FREE(inst->ztx);
```
dovholuknf, line 691: no `ZITI_LOG` in this situation? presumably because `set_tnlr_options` does any logggin?

dovholuknf (approved):
two small things but otherwise lgtm

lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
scareything: yeah, unless the caller is aware that a ziti_context is allocated by init_ziti_instance and goes out of their way to free it, calling more than once with the same ziti_instance will leak ztx.

dovholuknf: i see. why not just return an error? if i call it twice and stored off that context thinking i was making a second one that might be 'surprising'?

PR #754, opened 2023-11-13
Resolve hostnames in uv worker threads

### LINKED ISSUE #753: dns queries for hostnames in hosted service addresses can block the uv loop
When a connection is made to a hosted service, hostnames in the address[es] field of the service configuration are looked up synchronously on the uv loop. This means that packet reading /writing from the tun, other libuv socket activity, and really anything else that the tsdk would otherwise do will be blocked for the duration of the dns lookup.

libuv provides an async DNS query that should be used for hosted server address lookups.

scareything:
fixes #753

### DIFF
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -1918,6 +1918,7 @@ static void interrupt_handler(int sig) {
 #endif
 
 static void run(int argc, char *argv[]) {
+    setenv("UV_THREADPOOL_SIZE", "8", 0);
     uv_loop_t *ziti_loop = uv_default_loop();
     main_ziti_loop = ziti_loop;
     uv_cond_init(&stop_cond);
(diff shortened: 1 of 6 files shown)


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1918,6 +1918,7 @@ static void interrupt_handler(int sig) {
 #endif
 
 static void run(int argc, char *argv[]) {
+    setenv("UV_THREADPOOL_SIZE", "8", 0);
```
ekoby, line 1921: I would not fix that here, let it be set at deployment

scareything: I put this here to effectively increase the default to something reasonable for folks who run from the command line. The last argument to `setenv` prevents the call from doing anything if the variable is already present in the environment.

ekoby: realistically if DNS resolution is slow, it can fill an 8-worker pool just as well as a 4-worker one

ekoby (approved):

PR #759, opened 2023-11-21
update vcpkg baseline used in CI package workflow

scareything:


lib/ziti-tunnel-cbs/include/ziti/ziti_dns.h
```diff
@@ -20,6 +20,7 @@
 #include <ziti/ziti_tunnel.h>
 
 #define DNS_NO_ERROR 0
+#define DNS_SERVFAIL 2
```
qrkourier, line 23: Will this change how ZET's nameserver responds to hits or misses?

qrkourier (approved):
It raised a question for me about the nameserver change, and I see this fixes the CI package workflow.

lib/ziti-tunnel-cbs/include/ziti/ziti_dns.h
scareything: Actually this change is benign as no code uses the SERVFAIL symbol. I didn't intend to include this file in the comment, and it will be backed out shortly.

PR #771, opened 2023-12-20
X86 windows build

dovholuknf:


scareything (approved):
So it turns out that we don't want windows-x86 for zet releases, but we made some solid improvements to the windows build scripting (wintun.dll is now dropped into the right place) and maybe x86 will be useful to somebody out there.

.github/workflows/cmake.yml
```diff
@@ -34,10 +34,6 @@ jobs:
             name: Windows x86_64
             preset: windows-x64
 
-#          - os: windows-latest
```
ekoby, line 37: why not enabling it?

toolchains/Windows-x86.cmake
```diff
@@ -0,0 +1,5 @@
+# cross-compile for windows/arm64 on windows/x64 host with visual studio
```
ekoby, line 1: should be `... for windows/x86`

.github/workflows/cmake.yml
scareything: Two reasons really:

1. Nobody is asking for it. We initially did this because we thought we needed it for the zdew builds. But it turns out that zdew has been packaging 64-bit zet since the beginning.
2. Maybe the build will be useful to somebody out there with a special need, but including it in our releases might lead to more confusion than anything else.

toolchains/Windows-x86.cmake
scareything: And in fact this toolchain can be used on a Windows/arm64 host

PR #772, opened 2023-12-20
log tcp src and dst in LOG_STATE. also log tcp flags of incoming segments

scareything:
`LOG_STATE` now looks like this:
```
VERBOSE tunnel-sdk:tunnel_tcp.c:141 on_tcp_client_data() status 0 src[tcp:100.64.1.1:64905] dst[tcp:100.64.1.5:22] state[4/ESTABLISHED] flags[0x101] service[ssh]
```
incoming segments now look like this:
```
TRACE tunnel-sdk:tunnel_tcp.c:366 recv_tcp() received segment src[tcp:100.64.1.1:64905] dst[tcp:100.64.1.5:22] flags[PSH,ACK]
```

### DIFF
--- a/lib/ziti-tunnel/tunnel_tcp.c
+++ b/lib/ziti-tunnel/tunnel_tcp.c
@@ -25,8 +25,15 @@
 #define MIN(a,b) ((a)<(b) ? (a) : (b))
 #endif
 
-#define LOG_STATE(level, op, pcb, ...) \
-TNL_LOG(level, op " %p, state=%d(%s) flags=%#0x", ##__VA_ARGS__, pcb, pcb->state, tcp_state_str(pcb->state), pcb->flags)
+#define LOG_STATE(level, op, pcb, ...) do { \
+    io_ctx_t *io = ((io_ctx_t *)(pcb)->callback_arg); \
+    tunneler_io_context tnlr_io = io ? io->tnlr_io : NULL; \
+    const char *service_name = tnlr_io ? tnlr_io->service_name : ""; \
+    TNL_LOG(level, op " src[%s] dst[%s] state[%d/%s] flags[%#0x] service[%s]", ##__VA_ARGS__, \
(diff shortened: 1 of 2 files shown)


lib/ziti-tunnel/tunnel_tcp.c
```diff
@@ -339,12 +346,25 @@ u8_t recv_tcp(void *tnlr_ctx_arg, struct raw_pcb *pcb, struct pbuf *p, const ip_
     struct tcp_hdr *tcphdr = (struct tcp_hdr *)((char*)p->payload + iphdr_hlen);
     u16_t src_p = lwip_ntohs(tcphdr->src);
     u16_t dst_p = lwip_ntohs(tcphdr->dest);
+    char src_str[IPADDR_STRLEN_MAX];
```
ekoby, line 349: should this be bigger to include the port?

lib/ziti-tunnel/tunnel_tcp.c
```diff
@@ -339,12 +346,25 @@ u8_t recv_tcp(void *tnlr_ctx_arg, struct raw_pcb *pcb, struct pbuf *p, const ip_
...
     char dst_str[IPADDR_STRLEN_MAX];
+    ipaddr_ntoa_r(&src, src_str, sizeof(src_str));
     ipaddr_ntoa_r(&dst, dst_str, sizeof(dst_str));
-    TNL_LOG(TRACE, "received segment %s:%d->%s:%d",
-            ipaddr_ntoa(&src), src_p, dst_str, dst_p);
-
     u8_t flags = TCPH_FLAGS(tcphdr);
+    char flags_str[40] = { 0 };
+
+    if (flags & TCP_FIN) strcat(flags_str, "FIN,");
```
ekoby, line 356: this seem like it would be slow to do that for every packet even if it is not logged (when log level < TRACE)

ekoby (approved):

PR #774, opened 2023-12-22
only proxy dns queries for supported query types

scareything:


lib/ziti-tunnel-cbs/ziti_dns.c
```diff
@@ -701,12 +703,12 @@ ssize_t on_dns_req(const void *ziti_io_ctx, void *write_ctx, const void *q_packe
...
     if (q->type == NS_T_A || q->type == NS_T_AAAA) {
         process_host_req(req);
-    } else {
+    } else if (req->msg.recursive) {
         // find domain requires normalized name
         char reqname[MAX_DNS_NAME];
         check_name(q->name, reqname, NULL);
         dns_domain_t *domain = find_domain(reqname);
-        if (domain) {
+        if (domain && (q->type == NS_T_MX || q->type == NS_T_TXT || q->type == NS_T_SRV)) {
```
ekoby, line 711: i think this check needs to be inside `if (domain)` clause

otherwise request leaks to outside resolver

lib/ziti-tunnel-cbs/ziti_dns.c
```diff
@@ -701,12 +703,12 @@ ssize_t on_dns_req(const void *ziti_io_ctx, void *write_ctx, const void *q_packe
 
     if (q->type == NS_T_A || q->type == NS_T_AAAA) {
         process_host_req(req);
-    } else {
+    } else if (req->msg.recursive) {
```
ekoby, line 706: I worry this may break windows domain stuff. maybe only do this check for upstream?

lib/ziti-tunnel-cbs/ziti_dns.c
```diff
@@ -700,14 +702,16 @@ ssize_t on_dns_req(const void *ziti_io_ctx, void *write_ctx, const void *q_packe
...
     } else {
         // find domain requires normalized name
         char reqname[MAX_DNS_NAME];
         check_name(q->name, reqname, NULL);
         dns_domain_t *domain = find_domain(reqname);
         if (domain) {
-            proxy_domain_req(req, domain);
+            if (q->type == NS_T_MX || q->type == NS_T_TXT || q->type == NS_T_SRV) {
+                proxy_domain_req(req, domain);
+            }
```
ekoby, line 714: Where do we respond if it is not a supported query?

scareything: We haven't in the past. A previous commit on this PR fixed that, but then I had second thoughts because resolvers can be touchy and I worried about changing any behavior outside of filtering the queries that are proxied.

I'll take another shot at ensuring completion, but the logic is getting a little twisted here.

ekoby (approved):

scareything:
fixes #776

PR #777, opened 2023-12-26
complete proxy dns request when ziti_write fails

scareything:


lib/ziti-tunnel-cbs/ziti_dns.c
```diff
@@ -639,20 +655,23 @@ static void proxy_domain_req(struct dns_req *req, dns_domain_t *domain) {
...
+        int rc = ziti_write(domain->resolv_proxy, wr->json, jsonlen, on_proxy_write, wr);
         if (rc == ZITI_OK) {
             return;
         }
 
         ZITI_LOG(WARN, "failed to write proxy resolve request[%04x]: %s", req->id, ziti_errorstr(rc));
         ziti_close(domain->resolv_proxy, NULL);
         domain->resolv_proxy = NULL;
-        free(json);
+        free(wr->json);
```
dovholuknf, line 673: this sure seems like a double free to me, but it must not be somehow. wouldn't the on_proxy_write free these?

scareything: The write callback won’t get called if ziti_write returns error.

dovholuknf: aaaahhhhh... ok. that makes more sense but wasn't obvious.

dovholuknf (approved):

PR #778, opened 2024-01-04
add a script to create a debug tarball

qrkourier:


qrkourier:
Should I filter out any of those system calls in the *.strace file, or are they all potentially useful?

qrkourier:
```bash
 ❯ sudo bash ./scripts/debug.bash         
INFO: debug bundle created at /tmp/ziti-edge-tunnel-0.22.20-local-2024-01-04T21:54Z.tgz from files in /tmp/tmp.COFIjI3FXE
```

scripts/debug.bash
```diff
@@ -0,0 +1,148 @@
...
+    ) &> "${HOST_INFO_FILE}"
+    echo -n "."
+    
+    # save the current service unit invocation's log messages
+    journalctl _SYSTEMD_INVOCATION_ID="$(systemctl show -p InvocationID --value ziti-edge-tunnel.service)" -l --no-pager \
+    &> "${LOG_FILE}"
+    echo -n "."
+    
+    # save the threads and backtrace
+    timeout --kill-after=1s 3s \
```
scareything, line 80: `backtrace` is only going to show the stack for a specific thread, and probably not the most interesting one. we can get a backtrace for all threads with this gdb command:
```
thread apply all backtrace
```

scripts/debug.bash
```diff
@@ -0,0 +1,148 @@
...
+        exit 1
+    fi
+    
+    # save the systemd-resolved state
+    (
+        set +e
+        set -x
+        systemctl status systemd-resolved.service
+        resolvectl status
+        resolvectl dns ziti0
```
scareything, line 57: fyi the tun device might not be `ziti0`

scareything:
It's be good to get a few more tidbits, like `ip link show`, `ip addr show`, `ip route show`

scripts/debug.bash
```diff
@@ -0,0 +1,148 @@
...
+    # save the threads and backtrace
+    timeout --kill-after=1s 3s \
+        gdb /opt/openziti/bin/ziti-edge-tunnel \
+            --pid "${ZET_PID}" \
+            --batch \
+            --ex "set verbose on" \
+            --ex "set pagination off" \
+            --ex "info threads" \
+            --ex "backtrace" \
+            --ex "quit" \
```
scareything, line 88: I'd expect the `quit` command to cause a prompt. Does it? Either way it's probably safest to `--ex "detach"` first.

qrkourier: Correct, `--batch` was necessary to avoid a hung `gdb` command that attaches.

scripts/debug.bash
qrkourier: I'll stop specifying the device for both of those commands.

qrkourier:
@scareything Thanks for those refinements. It's ready for another review.

dovholuknf (approved):

PR #787, opened 2024-01-24
grant read on ziti identities to members of group 'ziti'

qrkourier:


dovholuknf (approved):

qrkourier:
I proposed this change because a typical Linux user who has taken the manual step of adding themselves to the `ziti` permission group so they can send IPC commands like `ziti-edge-tunnel add` or `submit_mfa` is still unable to read the identity files that were added.

I found myself frequently adjusting the permissions on those files with root permission, i.e., `sudo chmod -Rc g+rX /opt/openziti/etc/identities`, and estimated that other users too are likely to expect read access after manually adding themselves to the permission group.

PR #794, opened 2024-02-08
document cross-compiling ZET and DEB

### LINKED ISSUE #792: document cross-building the Linux package and Linux binary
Document the cross-building tools we have for the Linux package (distro-specific) and Linux binary (generic for libc 2.27) so that developers can cross-build conveniently, end users can leverage features unavailable in the release build, etc.

qrkourier:
resolves #792

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -182,7 +182,7 @@ docker run \
   --volume "${PWD}:/github/workspace" \
   --workdir "/github/workspace" \
   --env "TLSUV_TLSLIB=openssl" \
-  bionic-builder \
+  jammy-builder \
     ci-linux-arm64
 ```
 
@@ -190,7 +190,7 @@ docker run \
 
 All of the Ziti projects that leverage Ziti's C-SDK are built with a shared builder image: `openziti/ziti-builder`. This
 project provides a wrapper script for cross-building the generic `ziti-edge-tunnel` binary using this builder image
-optimized for compatibility, i.e., libc 2.27 and static TLS library.
+optimized for compatibility, i.e., libc 2.27 and static Mbed-TLS library.
 
 Without any arguments, the `ziti-builder.sh` script will build the `bundle` target with the `ci-linux-x64` (amd64)
 preset, placing the resulting ZIP archive in `./build/bundle/`, and the bare executable in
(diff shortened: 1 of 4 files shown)


CMakePresets.json
```diff
@@ -184,15 +184,24 @@
     },
     {
       "name": "ci-linux-x64-static-libssl",
-      "inherits": [ "ci-linux-static-libssl", "ci-linux-x64" ]
+      "inherits": [
```
qrkourier, line 187: these were pretty printed by `jq`, so I conceded to leave them that way because I anticipate continually modifying this JSON file in scripts until the Mbed-TLS vs. OpenSSL flapping is settled.

scripts/openssl-build.bash
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
qrkourier, line 1: This documents the workaround for building with OpenSSL and is referenced by BUILD.md.

scripts/switchtls-build.bash
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
scareything, line 1: instead of modifying the checked-in CMakePresets file, I think it would be better to mention `TLSUV_TLSLIB` and show how to set it in CMakeUserPresets.json.

qrkourier:
This branch's BUILD.md intends to use this change to the generic builder image: https://github.com/openziti/ziti-builder/pull/31

BUILD.md
```diff
@@ -143,29 +143,67 @@ the number of jobs to use, which should ideally be specified to the number of
...
 
-To build for a specific target architecture, use the `-p` argument to specify the vcpkg preset.
+#### Run the Package Builder Container
+
+Cross-build the Debian package for arm64 in the x64 builder container. The `ci-linux-arm64` in this example is an
+architecture-specific CMake [preset][1], and the optional TLS library variable overrides the default library, MBed-TLS.
 
 ```bash
-./ziti-builder.sh -p ci-linux-arm64
+docker run \
```
scareything, line 180: does this docker run command require `--platform linux/amd64` if run from an arm64 host?

scareything (approved):

PR #796, opened 2024-02-09
statically build in openssl instead of mbedtls

qrkourier:


qrkourier:
In addition to the differences in #707, we'd need to revert the vcpkg.jso dependency to `openssl` as was the case [in that branch at the time](https://github.com/openziti/ziti-tunnel-sdk-c/blob/88621a937eee78bb8899ad35ac2bc592a6f13ce5/vcpkg.json#L11).

The vcpk.json file currently declares `mbedtls` only.

qrkourier:
superseded by https://github.com/openziti/ziti-tunnel-sdk-c/pull/879

Issue #517, opened 2022-10-13
ziti-edge-tunnel enroll "COULD_NOT_PROCESS_CSR" on Red Hat 9 / Rocky 9


qrkourier:
~~I tried the 0.20.0 release binary and~~ I built `ziti-edge-tunnel` from source tag v0.20.0 on Rocky 9 and the `enroll` command always gets this error:
> COULD_NOT_PROCESS_CSR

I was able to enroll with the same binary artifact running on another Linux system (not Red Hat 9) without encountering this error, and I was able to enroll the same JWT with the release binary 0.20.0 running on another Linux system, so it doesn't seem to be a problem with the controller or the JWT or the binary itself, except when it's running on Red Hat 9.

The release binary is built with Mbed-TLS, and I set USE_OPENSSL=ON when I built from source, so both TLS implementations have been tried.

```bash
[rocky@ip-172-31-4-195 ~]$ sudo ./ziti-edge-tunnel enroll -j /opt/openziti/etc/identities/rh9client2.jwt -i /opt/openziti/etc/identities/rh9client2.json
(15357)[        0.520]   ERROR ziti-sdk:ziti_enroll.c:227 enroll_cb() failed to enroll with controller: https://7ce7e424-6a92-4ff2-9459-ebbba32346fa.production.netfoundry.io:443 COULD_NOT_PROCESS_CSR (The supplied csr could not be processed)
(15357)[        0.520]   ERROR ziti-edge-tunnel:ziti-edge-tunnel.c:1992 enroll_cb() enrollment failed: COULD_NOT_PROCESS_CSR(-3)

[rocky@ip-172-31-4-195 ~]$ ldd ./ziti-edge-tunnel
        linux-vdso.so.1 (0x00007fffab524000)
        libssl.so.3 => /lib64/libssl.so.3 (0x00007f5fa08ad000)
        libcrypto.so.3 => /lib64/libcrypto.so.3 (0x00007f5fa0481000)
        libatomic.so.1 => /lib64/libatomic.so.1 (0x00007f5fa0478000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f5fa039d000)
        libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f5fa0389000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f5fa017d000)
        libz.so.1 => /lib64/libz.so.1 (0x00007f5fa0163000)
[...]

PR #799, opened 2024-02-15
enable building rh9 rpm with shared openssl

qrkourier:


qrkourier:
Clarification: this change makes it possible for us to test the OpenSSL build using the CPack builder machinery. This does not change the release build to OpenSSL from static Mbed-TLS.

.github/actions/openziti-tunnel-build-action/redhat-9/entrypoint.sh
```diff
@@ -45,6 +45,7 @@ done
         --preset "${cmake_preset}" \
         -DCMAKE_BUILD_TYPE="${cmake_config}" \
         -DBUILD_DIST_PACKAGES=ON \
+        "${TLSUV_TLSLIB:+-DTLSUV_TLSLIB=${TLSUV_TLSLIB}}" \
```
scareything, line 48: Why do this for redhat-9 only? If this is just a quick trick for us then maybe a comment here to explain the purpose.

Is there any value in doing it for other distros (keeping in mind that just setting TLSUV_TLSLIB to openssl is only sufficient on distros that already have libssl3 - other distros will also need "openssl" to be added to vcpkg.json). 

There is one downside to this and I'm not sure how to handle it. "mbedtls" is currently defined as a dependency in vcpkg.json, which means that vcpkg will download and build mbedtls whether or not the build actually uses it (which is determined by TLSUV_TLSLIB).

scareything (approved):

.github/actions/openziti-tunnel-build-action/redhat-9/Dockerfile
```diff
@@ -1,8 +1,10 @@
 ARG CMAKE_VERSION="3.26.3"
+ARG VCPKG_VERSION="2024.01.12"
```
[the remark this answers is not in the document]
qrkourier: I see what you mean. I'd like to converge on a pattern we can use for the CPack cross-builder images as well as the generic ziti-builder image (cross-builder for everything else that uses the C-SDK that's not a ZET Linux package). I'll keep the trunk checkout idea in mind.

.github/actions/openziti-tunnel-build-action/redhat-9/entrypoint.sh
qrkourier: I extended the TLS lib override to RedHat9 so I could try to reproduce the enrollment problem in https://github.com/openziti/ziti-tunnel-sdk-c/issues/517#issuecomment-1947494141.

I see what you mean about building Mbed-TLS unnecessarily. I see how to temporarily remove it from vcpkg's deps list, or to replace it with "openssl" for building it statically for distros with libssl v1.

PR #801, opened 2024-02-19
fix mkdir command that enables clobbering tmp identity in container

qrkourier:


docker/docker-entrypoint.sh
```diff
@@ -61,7 +61,8 @@ fi
 # if identity env var is defined then do not look for mounted identities
 if [[ -n "${ZITI_IDENTITY_JSON:-}" ]]; then
     IDENTITIES_DIR="/tmp/openziti"
-    mkdir -m0700 "${IDENTITIES_DIR}"
+    # shellcheck disable=SC2174
+    mkdir -pm0700 "${IDENTITIES_DIR}"
```
scareything, line 65: What's this fixing? Is the /tmp directory actually missing sometimes? If so, how does that happen? We should prevent that if possible because /tmp is an important directory.

qrkourier: The problem is that `/tmp/openziti` already exists if a container is resumed, so `mkdir` always fails in that case with an error like "file exists."

The fix is to tolerate the existing directory because the goal of this hunk is to clobber `/tmp/openizti/${ZITI_IDENTITY_BASENAME}.json` with the value of `$ZITI_IDENTITY_JSON`.

scareything (approved):

PR #804, opened 2024-02-20
add group ziti to container images

qrkourier:


docker/Dockerfile.base
```diff
@@ -40,7 +40,7 @@ RUN mkdir -m0755 /licenses
 COPY ./LICENSE-Apache /licenses/apache.txt
 
 ### Add necessary Red Hat repos and packages
-RUN INSTALL_PKGS="iproute procps" && \
+RUN INSTALL_PKGS="iproute procps shadow-utils" && \
```
scareything, line 43: I think our security posture is reduced if we leave the shadow-utils package installed for use at runtime. Could we uninstall it after creating the group?

docker/Dockerfile.base
```diff
@@ -40,7 +40,7 @@ RUN mkdir -m0755 /licenses
...
     microdnf -y update --setopt=install_weak_deps=0 --setopt=tsflags=nodocs && \
     microdnf -y install --setopt=install_weak_deps=0 --setopt=tsflags=nodocs ${INSTALL_PKGS}
```
scareything, line 45: I guess `shadow-utils` should only be removed if it isn't in `${INSTALL_PKGS}`. Alternatively we could install in in the `fetch-ziti-artifacts` stage and `COPY --from=fetch-ziti-artifacts` the modified group files (/etc/group, /etc/gshadow) from the stage.

scareything (approved):

PR #824, opened 2024-03-08
enable setting only the enroll token var in default run state

qrkourier:


qrkourier:
This change minimizes the Docker run command by providing a default value `ZITI_IDENTITY_BASENAME=ziti_id`, e.g.,

```bash
docker run --env ZITI_ENROLL_TOKEN=$(< ./hello-host.jwt) openziti/ziti-host
```

scareything (approved):

PR #831, opened 2024-04-12
check dns response buffer as it is filled, and set truncate flag accordingly

scareything:


dovholuknf (approved):

lib/ziti-tunnel-cbs/dns_host.c
```diff
@@ -145,14 +145,48 @@ static int fmt_txt(const ns_msg *msg, const ns_rr* rr, dns_answer *ans, size_t m
...
-void do_query(const dns_question *q, dns_message *resp, resolver_t *resolver) {
-    uint8_t resp_msg[PACKETSZ];
-    int rc = res_nquery(resolver, q->name, ns_c_in, q->type, resp_msg, PACKETSZ);
+static uint8_t *send_and_parse_query(const dns_question *q, int class, ns_msg *ans, resolver_t *resolver) {
+    int buf_sz = PACKETSZ;
+    uint8_t *resp_msg = NULL;
+    int rc;
+
+    for (int i = 0; i < 2; i++) {
+        resp_msg = calloc(buf_sz, sizeof(uint8_t));
```
dovholuknf, line 154: probably could use a comment here wrt what `for (int i = 0; i < 2; i++) {` is all about

PR #838, opened 2024-04-29
build with gcc 14

scareything:


qrkourier (approved):
Do we need any ziti-builder image updates for this?

PR #845, opened 2024-06-03
support using http proxy for outbound connections

scareything:


lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -841,6 +878,26 @@ host_ctx_t *ziti_sdk_c_host(void *ziti_ctx, uv_loop_t *loop, const char *service
                 memcpy(&a->za, allowed_src_addrs[i], sizeof(a->za));
                 STAILQ_INSERT_TAIL(&host_ctx->allowed_source_addresses, a, entries);
             }
+
+            if (host_v1_cfg->proxy.type == ziti_proxy_server_type_http) {
+                const char *addr = host_v1_cfg->proxy.address;
+                if (addr != NULL && addr[0] != '\0') {
```
ekoby, line 884: you can use tlsuv_parse_url() here

lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -224,24 +245,27 @@ static void on_hosted_tcp_server_connect_complete(uv_connect_t *c, int status) {
...
+    if (status != 0) {
+        ZITI_LOG(ERROR, "proxy connect failed: %s (e=%d)", uv_strerror(status), status);
+        hosted_server_close(io);
+        return;
     }
-    
-    ziti_accept(io_ctx->client, on_hosted_client_connect_complete, NULL);
-    free(c);
+
+    uv_tcp_open(&io->server.tcp, sock);
```
ekoby, line 264: check return from uv_tcp_open() for errors

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1822,12 +1823,65 @@ static struct option run_host_options[] = {
 static const char* dns_upstream = NULL;
 static bool host_only = false;
 
+#include "tlsuv/http.h"
+static struct tlsuv_url_s *proxy_url = NULL;
```
ekoby, line 1827: why allocate this instead of static memory?

scareything: when it's null it means that no proxy was specified. alternatively there could be a separate boolean, but allocating it and checking for non-null felt cleaner to me.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -2952,6 +3014,7 @@ static int add_identity_opts(int argc, char *argv[]) {
 static CommandLine enroll_cmd = make_command("enroll", "enroll Ziti identity",
         "-j|--jwt <enrollment token> -i|--identity <identity> [-k|--key <private_key> [-c|--cert <certificate>]] [-n|--name <name>]",
         "\t-j|--jwt\tenrollment token file\n"
+        "\t-x|--proxy [username[:password]@]hostname_or_ip:port\tproxy to use when connecting to OpenZiti controller"
```
ekoby, line 3017: we should probably change this to allow other proxy types in the future (socks, socks5)
add `proto://` in front
then you can check proto to select proxy implementation (error/reject anything other than `http` now)

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
ekoby: you can just check proxy_url.host for non-NULL
if it wasn't specified it won't be set

also might avoid global var altogether since it is not needed after you set global connector

scareything: the init logic has been changed to use the global connector

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
scareything: done

ekoby (approved):
awesome

PR #854, opened 2024-06-12
eol xenial, bionic, and redhat7; adopt focal builder and node 20 actions

qrkourier:


qrkourier:
Partially redundant to https://github.com/openziti/ziti-tunnel-sdk-c/pull/812 and this can't merge until we solve https://github.com/openziti/ziti-builder/issues/26

scripts/ziti-builder.sh
```diff
@@ -125,10 +126,10 @@ function main() {
             -E make_directory \
             ./build  
         cmake \
-            --preset "${CMAKE_PRESET:-ci-linux-x64}" \
+            --preset "${CMAKE_PRESET:-ci-linux-x64-static-libssl}" \
```
scareything, line 129: Do we want this? I would think we want to let the build use the system's shared libssl.so if someone is building their own binary.

vcpkg-overlays/linux-syslibs/ubuntu20/zlib/portfile.cmake
```diff
@@ -0,0 +1 @@
+set(VCPKG_POLICY_EMPTY_PACKAGE enabled)
```
scareything, line 1: Can ubuntu 20 use the default openssl port? The intent of this port directory structure is for distros with sufficient package versions to use "default" - otherwise we end up making a new port for each and every distro that we build on.

scareything (approved):

qrkourier:
superseded by https://github.com/openziti/ziti-tunnel-sdk-c/pull/893

PR #892, opened 2024-07-09
Linux: Manage routes in a routing table dedicated to ZET.

tomc-xsec:
Linux has the capability to support multiple routing tables. So let's manage ZET routes in a dedicated routing table, distinct from main and default. The routing policy database is configured to prefer the routes managed in the ZET table over those stored in the main and default routing tables. This approach provides isolation, and thus additional security for ZET routes.

This serves two purposes:
- It acts as a safeguard against TunnelVision (CVE-2024-3661). For more information, visit: https://www.tunnelvisionbug.com/
- It simplifies application of multipathing and MPTCP
- Improves support for jumping between networks

### DIFF
--- a/lib/ziti-tunnel-cbs/CMakeLists.txt
+++ b/lib/ziti-tunnel-cbs/CMakeLists.txt
@@ -8,6 +8,7 @@ target_compile_definitions(ziti-tunnel-sdk-c
 add_library(ziti-tunnel-cbs-c STATIC
         ziti_tunnel_cbs.c
         ziti_hosting.c
+        ziti_hosting_socket.c
         ziti_tunnel_ctrl.c
         ziti_instance.h
         ziti_dns.c
--- a/lib/ziti-tunnel-cbs/ziti_hosting.c
+++ b/lib/ziti-tunnel-cbs/ziti_hosting.c
@@ -38,6 +38,15 @@
 
 #define KEEPALIVE_DELAY 60
 
+#ifndef WIN32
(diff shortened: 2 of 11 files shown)


tomc797:
This change has ZET preferring to manage routes in route table 90 (the decimal code for 'Z'). Rules are configured within the route policy database to utilize the ZET routing table. The rule has preference 23141 (The decimal code for 'Ze'). Excluded routes are managed using the 'throw' route type. When a matching throw route is encountered, the route lookup process immediately terminates, and the routing process proceeds to the next policy rule with lower priority.

tomc797:
Manage ziti-edge-tunnel routes in route table 90 (the decimal code for 'Z'). Rules are configured within the route policy database to utilize the route table 90. The rule has preference 23141 (The decimal code for 'Ze') and configured to select on traffic without the bypass fwmark. If sockets are configured with bypass mark, such as the sockets used to communicate with the controller, edge routers, and `hosted`, they bypass this lookup. In these cases, the traffic originating from these sockets will utilize the `main` and `default` routing tables. The added benefit is that if you jump networks (jump from WiFi to cell, for instance), there isn't a route that may interfere with reestablishing the connections.

tomc797:
Would it be possible to get a review? I would appreciate feedback.

scareything (commented):
Hi and thanks for contributing! I've skimmed the hosting portion of your changes, and I'll make another pass for the routing changes in the next day or so.

PR #907, opened 2024-07-19
Simpler IPC

ekoby:
simplifies IPC client cli subcommands (tunnel_status, etc) -- simple command/response client does not need an event loop
improves handling of incoming commands - use JSON to determine command boundaries


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -654,49 +645,55 @@ static void process_ipc_command(uv_stream_t *s, ssize_t len, char* base) {
...
-        queue_ipc_command(ipc_cmd_ctx, len, b->base);
-        if (b->base[len-1] == LAST_CHAR_IPC_CMD) {
-            char* new_buff = ipc_cmd_buffered(ipc_cmd_ctx);
-            ZITI_LOG(TRACE, "buffered cmd <%.*s>", (int) (strlen(new_buff) + 1), new_buff);
-            process_ipc_command(s, strlen(new_buff) + 1, new_buff);
-            free(new_buff);
+        json_tokener *parser = s->data;
+
+        size_t processed = 0;
+        while (processed < len) {
```
dovholuknf, line 667: Shouldn't this have some kind of escape-hatch max value? Something like a 1mb or 10 mb? something 'crazy' but not too crazy?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -2386,92 +2344,98 @@ static int dump_opts(int argc, char *argv[]) {
...
+#if _WIN32
+    HANDLE cmd_soc = CreateFileA(sockfile,
+                                 GENERIC_READ | GENERIC_WRITE,
+                                 0, NULL,
+                                 OPEN_EXISTING,
+                                 FILE_FLAG_OVERLAPPED, NULL);
+    if (cmd_soc == INVALID_HANDLE_VALUE) {
+        DWORD err = GetLastError();
+        fprintf(stderr, "failed to connect to pipe: %lu", err);
+        exit(1);
```
dovholuknf, line 2357: are you certain about this particular change with respect to `exit`-ing? Why is windows the only one that would end up in this situation? Shouldn't the other OS's all exit too in these situations?

This function is much harder to review with all these defines. the old code appears much simpler (just looking at it here, without the IDE) why were these changes needed?

ekoby: other OS's would also exit -- just a bit later -- after the write attempt. I'll add the exit on connect

there were multiple callback needed for a simple request/response pattern. which is definitely not simpler

also almost all `ziti-edge-tunnel <subcmd>` were not working on windows because command parsing was relying on the `'\0'/'\n'`

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
ekoby: previous implementation was not correct and would definitely crash if you send over 16K of data

dovholuknf (approved):
i still think we should prevent reading from the IPC 'forever' but other than that, lgtm

PR #925, opened 2024-08-08
implement tunnel enroll command

ekoby:


lib/ziti-tunnel-cbs/ziti_tunnel_ctrl.c
```diff
@@ -626,6 +627,32 @@ static int process_cmd(const tunnel_command *cmd, command_cb cb, void *ctx) {
             break;
         }
 
+        case TunnelCommand_Enroll: {
```
dovholuknf, line 630: This needs to work with `ziti-edge-tunnel.c`'s `TunnelCommand_AddIdentity`. Is that coming?

ekoby: > This needs to work with `ziti-edge-tunnel.c`'s `TunnelCommand_AddIdentity`. Is that coming?

no, this command does enrollment and returns `ziti_config`, `AddIdentity` makes assumptions on where/how config is stored (file system) which is not true for other consumers of tunneler SDK

ideally AddIdentity should be re-implemented as Enroll -> store -> LoadIdentity, but it is out of scope now

PR #926, opened 2024-08-09
get the build_date a better way using builtin cmake command

dovholuknf:


ekoby:
it is better to use CMake built-in: `string(TIMESTAMP ...)` https://cmake.org/cmake/help/latest/command/string.html#timestamp

scareything (approved):

PR #955, opened 2024-08-27
upgrade cpack builders prior to depsolving

qrkourier:


qrkourier:
I made this change because I encountered an error where dnf in rockylinux:9 was unable to resolve dependency glibc = 2.34-100.el9_4.3 while trying to install glibc-devel in the CPack builder image.

almalinux:9 didn't have the same problem, so I switched the source for the RedHat9 CPack builder image to alma.

Along the way I noticed that we never upgrade the CPack builder images before depsolving or building. I added that because it seems more likely to yield newer, better build tools from the appstream, but I'm not married to the idea if the reviewer insists we should instead optimize for stability of the builder images. It's not a "service" image per se, only a "builder" image.

scareything (approved):

PR #958, opened 2024-08-29
use RelWithDebInfo build type

scareything:


qrkourier (changes requested):
I think we need a callable workflow input on https://github.com/openziti/ziti-tunnel-sdk-c/blob/add.debug.symbols/.github/workflows/cmake.yml to make this work with the release.yml workflow. That way, `config: Release` can be set when the release triggers this composite action to build the release artifacts.

scareything:
> I think we need a callable workflow input on https://github.com/openziti/ziti-tunnel-sdk-c/blob/add.debug.symbols/.github/workflows/cmake.yml to make this work with the release.yml workflow. That way, `config: Release` can be set when the release triggers this composite action to build the release artifacts.

`config` _is_ an input to the action. It turns out we don't use it in our workflows though, so this PR changes the default `config` from "Release" to "RelWithDebInfo".  So anyone who wants to build pure release is able to do so.

We want to add debug symbols (symbol names only) in our released binaries to facilitate debugging issues that arise in the field.

qrkourier (approved):

PR #977, opened 2024-09-19
remove dns client's active requests from ziti_dns.requests when closing client

scareything:


dovholuknf (approved):

lib/ziti-tunnel-cbs/ziti_dns.c
```diff
@@ -287,10 +287,19 @@ void* on_dns_client(const void *app_intercept_ctx, io_ctx_t *io) {
     return clt;
 }
 
+static void remove_dns_req(void *p) {
+    struct dns_req *req = p;
```
ekoby, line 291: should it be freed? it is removed from both maps

ekoby (approved):

PR #1000, opened 2024-10-04
Rc add zfw calls

r-caamano:
This pull request adds system calls to zfw for the purpose allowing bi-directional transparency.   I allows user to specify diverter function either via command line argument or environmental variable with command line taking precedence.  ```-D, diverter <comma separated interface list> or ZITI_DIVERTER =<comma separated interface list>.   I also added a more restrictive option which in addition to transparency support also provides full firewall filtering -f, --diverter-fw <interface list> | DIVERTER_FIREWALL=<interface list>.  It assumes zfw binaries/ebpf objects are in /opt/openziti/bin by default but can be modified via environmental variable ZFW_OBJECT_PATH=<path>.

### DIFF
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -2696,7 +2696,7 @@ void setup_xdp(char *tun_name){
         int status =0;
         if(!(waitpid(pid, &status, 0) < 0)){
             if(!(WIFEXITED(status) && !WEXITSTATUS(status))){
-                printf("diverter xdp tunnel redirect not set on dev %s\n","ziti0");
+                printf("diverter xdp tunnel redirect not set on dev %s\n", tun_name);
             }
         }
     }
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -148,7 +148,7 @@ static char *configured_cidr = NULL;
 static char *configured_log_level = NULL;
 static char *configured_proxy = NULL;
 static char *config_dir = NULL;
-static char *diverterIf = NULL;
+static char *diverter_if = NULL;
 static bool diverter = false;
 static bool firewall = false; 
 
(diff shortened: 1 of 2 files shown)


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1996,6 +2160,772 @@ static void interrupt_handler(int sig) {
...
 #endif
 
+#if __linux__
+static void diverter_update(char *ip, char *mask, char *lowport, char *highport, char *protocol, char *service_id, char *action){
+    bool state = true;
+    unsigned short random_port = 0;
+    unsigned char count = 0;
+    int rndm;
+    uv_random(NULL, NULL, &rndm, sizeof(rndm), 0, NULL);
+    random_port = htons(1024 + rndm % (65535 - 1023));
```
scareything, line 2170: Is `htons` (network byte order) really needed here? I could see it if the port value was going into a packet that gets put on the wire and received by hosts of any architecture, but I think we're just passing a command line option to a local process?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1996,6 +2160,772 @@ static void interrupt_handler(int sig) {
...
+    int fd = open("/dev/null", O_WRONLY);
+    if (fd == -1){
+        return; 
+    }
+    dup2(fd, STDOUT_FILENO);
+    dup2(fd, STDERR_FILENO);
+    close(fd);
+    char *protocols[2] = {"tcp", "udp"};
+    for(int x = 0; x < 2; x++){
+        char *const parmList[] = {zfw_path, "-I", "-c", prefix, "-m", prefix_len, "-l", "1" , "-h", "65535", "-t", "65535", "-p", protocols[x], NULL};
```
scareything, line 2636: `prefix` is an int value here. should this be a string IP?

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
r-caamano: No its not.  I had copied this function over from where I needed to pass it directly to an ebpf map key which stores the port in network order but here it can be in host order, so the htons can be omitted.

scareything:
superseded by #1054

PR #1009, opened 2024-10-08
configure tun interface with ioctls

scareything:


scareything:
> Nice. This may mean we could remove the `iproute` package from [the container images](https://github.com/openziti/ziti-tunnel-sdk-c/blob/set.tun.addr.with.ioctl/docker/ziti-edge-tunnel.Dockerfile#L24C19-L24C26), but it doesn't seem necessary.

Not yet, but this does take us one step in that direction. We still issue `ip` commands for route creation.

PR #1031, opened 2024-10-28
use persisted config path on darwin

scareything:


dovholuknf (approved):
not the "app" folder? `~/Library/Group Containers/MN5S649TXM.ZitiPacketTunnel.group`

scareything:
> not the "app" folder? `~/Library/Group Containers/MN5S649TXM.ZitiPacketTunnel.group`

I think it might get confusing if zet plays in zde's sandbox.

PR #1034, opened 2024-10-30
revert previous until signal issues are understood

scareything:


ekoby (approved):
if anything we should put all that diverter code into a separate file

PR #1049, opened 2024-11-25
refinements for external auth support

dovholuknf:
various fixes, refinements while doing ext-auth support. brings in some changes from the other branch that was reviewed (but probably won't merge due to conflicts)
* checks for only one -I param supplied,
* saves the config file in the same location as the -I (config dir) now
* normalizes identities to lowercase on windows to avoid copy/paste issues
* change tunnel status to not save the config file (it was being written every 5s due to the ziti monitor on windows)
* clean up some warnings


programs/ziti-edge-tunnel/instance.c
```diff
@@ -881,6 +895,33 @@ void set_ziti_status(bool enabled, const char* identifier) {
...
+    } else {
+        discriminator = empty_delim; // can't use NULL so use the empty delim
+    }
+
+    size_t prefix_len = strlen(DEFAULT_EXECUTABLE_NAME);
+    size_t discrim_len = strlen(discriminator);
+    size_t delim_len = strlen(delim_to_use);
+
+    char *zet_instance_id = calloc(prefix_len + delim_len + discrim_len + 1, sizeof(char));
+    sprintf(zet_instance_id, "%s%s%s", DEFAULT_EXECUTABLE_NAME, delim_to_use, discriminator);
```
scareything, line 922: IMO this function would be easier to read if it didn't use `sprintf` for the NULL case. e.g.:
```
char *zet_instance_id = NULL;

if (discriminator) {
    zet_instance_id = calloc(strlen(DEFAULT_EXECUTABLE_NAME) + strlen(discriminator) + 2 /* separator + nul */, sizeof(char));
    sprintf(zet_instance_id, "%s.%s", DEFAULT_EXECUTABLE_NAME, discriminator);
} else {
    zet_instance_id = strdup(DEFAULT_EXECUTABLE_NAME);
}

return zet_instance_id;
```

programs/ziti-edge-tunnel/netif_driver/darwin/utun.c
```diff
@@ -283,4 +283,8 @@ netif_driver utun_open(char *error, size_t error_len, const char *cidr) {
         }
     }
     return driver;
+}
+
+const char* get_tun_name(netif_handle tun) {
```
scareything, line 288: could this be added to the `netif_driver` interface instead of a standalone function? e.g:

netif_driver.h
```
typedef int (*netif_close_cb)(netif_handle dev);
typedef const char * (*get_name_fn)(netif_handle dev);
// ... other typedefs

typedef struct netif_driver_s {
  get_name_fn get_name;
  // ... other fields
} netif_driver_t;
```

it would then be called through the netif driver pointer that was returned by `tun_open`. e.g.:

```
const char *tun_name = tun->get_name(tun->handle);
```

this would follow the pattern for all of the other implementation-specific functions, so for instance you put all of the wide chart stuff in the windows implementaion of tun_get_name instead of ifdefing it in ziti-edge-tunnel.c

programs/ziti-edge-tunnel/process_cmd.c
```diff
@@ -207,55 +248,109 @@ bool process_tunnel_commands(const tunnel_command *tnl_cmd, command_cb cb, void
...
+            tunnel_add_identity tunnel_add_identity_cmd = {0};
+            int parse_result = parse_tunnel_add_identity(&tunnel_add_identity_cmd, tnl_cmd->data, strlen(tnl_cmd->data));
+
+            if (parse_result < 0) {
+                result.error = "invalid command - could not parse";
                 result.success = false;
+                free_tunnel_add_identity(&tunnel_add_identity_cmd);
                 break;
             }
+            char *null = NULL;
```
scareything, line 272: why not use `NULL`

programs/ziti-edge-tunnel/process_cmd.c
```diff
@@ -207,55 +248,109 @@ bool process_tunnel_commands(const tunnel_command *tnl_cmd, command_cb cb, void
...
+            bool is_jwt = tunnel_add_identity_cmd.jwtFileName != null || tunnel_add_identity_cmd.jwtContent != null;
+            bool is_url = tunnel_add_identity_cmd.controllerURL != null;
+            bool is_3rd_party_ca = tunnel_add_identity_cmd.cert != null || tunnel_add_identity_cmd.key != null;
+            int enrollment_methods_supplied = (is_jwt + is_url + is_3rd_party_ca);
```
scareything, line 276: this would be clearer (and I think safer) if it was a bool:
```
bool enrollment_methods_supplied = is_jwt || is_url || is_3rd_party_ca;
```

programs/ziti-edge-tunnel/process_cmd.c
```diff
@@ -207,55 +248,109 @@ bool process_tunnel_commands(const tunnel_command *tnl_cmd, command_cb cb, void
...
 
-            if (tunnel_add_identity_cmd.jwtContent == NULL && tunnel_add_identity_cmd.controllerURL == NULL) {
-                result.error = "jwt content not provided";
+            if (enrollment_methods_supplied == 0) {
+                result.error = "no enrollment options detected. either JWT or URL must be specified.";
```
scareything, line 279: 3rd party ca is ok too?

programs/ziti-edge-tunnel/process_cmd.c
```diff
@@ -207,55 +248,109 @@ bool process_tunnel_commands(const tunnel_command *tnl_cmd, command_cb cb, void
...
             }
 
-            char* extension = strstr(tunnel_add_identity_cmd.jwtFileName, ".jwt");
-            size_t length;
-            if (extension != NULL) {
-                length = extension - tunnel_add_identity_cmd.jwtFileName;
+            char new_identifier_without_ext[FILENAME_MAX] = {0};
+            char new_identifier_with_ext[FILENAME_MAX] = {0};
+            char new_identifier_path[FILENAME_MAX] = {0};
+            if(is_jwt) {
```
scareything, line 288: whitespace

programs/ziti-edge-tunnel/process_cmd.c
```diff
@@ -207,55 +248,109 @@ bool process_tunnel_commands(const tunnel_command *tnl_cmd, command_cb cb, void
...
+                if (tunnel_add_identity_cmd.jwtFileName == NULL) {
```
scareything, line 289: Isn't it ok for either `jwtFileName` or `jwtContent` to be null?

programs/ziti-edge-tunnel/windows-scripts.c
```diff
@@ -302,15 +301,38 @@ void remove_nrpt_rules(uv_loop_t *nrpt_loop, model_map *hostnames) {
     }
 }
 
-void remove_all_nrpt_rules() {
+char* get_nrpt_comment(char* zet_id, bool exact) {
```
scareything, line 304: `zet_id` can be const. also I think you could drop the `exact` parameter and use `zet_id != NULL` instead.

programs/ziti-edge-tunnel/windows-scripts.c
```diff
@@ -302,15 +301,38 @@ void remove_nrpt_rules(uv_loop_t *nrpt_loop, model_map *hostnames) {
...
+        name = DEFAULT_EXECUTABLE_NAME;
+        format = "$_.Comment.StartsWith('Added by %s')";
+    }
+    size_t tot_chars = strlen(format) + strlen(name);
+    nrpt_filter = calloc(tot_chars + 1, sizeof(char));
+    snprintf(nrpt_filter, tot_chars, format, name);
+    return nrpt_filter;
+}
+
+void remove_all_nrpt_rules(char* zet_id, bool exact) {
```
scareything, line 322: same thought here as above for `get_nrpt_comment`

scareything (approved):
lets just add the get_name function to the netif_driver interface pls.

programs/ziti-edge-tunnel/instance.c
dovholuknf: 👍

programs/ziti-edge-tunnel/process_cmd.c
dovholuknf: first implementation needed it, but new/better impl doesn't. will make this change. thx

programs/ziti-edge-tunnel/process_cmd.c
dovholuknf: 3rd party ca requires a jwt, so it's not needed. i updated the boolean from above accordingly

programs/ziti-edge-tunnel/process_cmd.c
dovholuknf: no? this is during add identity, the first param is the filename to produce, the second is the content to use to enroll with. We can discuss tomorrow

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -367,12 +384,9 @@ static void load_identities(uv_work_t *wr) {
                 continue;
             }
 
-            if (strcasecmp(file.name, get_config_file_name(NULL)) == 0) {
+            if (ends_with(config_file, file.name)) {
```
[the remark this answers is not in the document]
dovholuknf: it _shouldn't_ matter, but i changed to use `strcasecmp` - good eyes as always! there's no basename on windows apparently so i'm fine with a bespoke func myself. another thing we can discuss tomorrow

PR #1054, opened 2024-12-03
Rc add zfw calls 1.2.x

scareything:
changes from #1000 applied to main instead of release-1.x

### DIFF
--- a/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
+++ b/programs/ziti-edge-tunnel/ziti-edge-tunnel.c
@@ -993,11 +993,6 @@ static int run_tunnel(uv_loop_t *ziti_loop, uint32_t tun_ip, uint32_t dns_ip, co
             exit(1);
         }
     }
-#else
-    if (diverter) {
-        ZITI_LOG(INFO,"Diverter features only supported on Linux");
-        exit(1);
-    }
 #endif
     run_tunneler_loop(ziti_loop);
     if (tun->close) {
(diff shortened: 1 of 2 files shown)


programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -78,6 +79,28 @@ static bool is_host_only();
 static void run_tunneler_loop(uv_loop_t* ziti_loop);
 static tunneler_context initialize_tunneler(netif_driver tun, uv_loop_t* ziti_loop);
 
+#if __linux__
+static void diverter_quit();
```
ekoby, line 83: we should make these into diverter interface

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -852,6 +925,61 @@ static int run_tunnel(uv_loop_t *ziti_loop, uint32_t tun_ip, uint32_t dns_ip, co
...
+        if(zifi_firewall && strlen(zifi_firewall)){
+            diverter = true;
+            firewall = true;
+            diverter_if = getenv("ZITI_FIREWALL");
+        }
+    }
+    char *diverter_env_path = getenv("ZFW_OBJECT_PATH");
+    if(diverter_env_path && strlen(diverter_env_path)){
+        diverter_path = diverter_env_path;
+        sprintf(zfw_path, "%s/%s", diverter_env_path, "zfw");
```
ekoby, line 944: use snprintf

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1264,6 +1415,129 @@ static void interrupt_handler(int sig) {
 }
 #endif
 
+#if __linux__
+// support interactions with zfw command line utility
+
+static void diverter_update(const char *ip, uint8_t prefix_len, uint16_t lowport, uint16_t highport, const char *protocol, const char *service_id, const char *action) {
+    int rndm;
+    uv_random(NULL, NULL, &rndm, sizeof(rndm), 0, NULL);
+    unsigned short random_port = 1024 + rndm % (65535 - 1023);
```
ekoby, line 1424: what happens if port is in use?

r-caamano: for zfw in tunnel mode it just cant be set to 0.  A setting of 0 indicates that a matching packet should be passed to the OS and not to directly forwarded to the ziti0 interface. As long as it between 1 and 65535 it does not matter if it is in use.  We could make it any static value between 1 and 65535 and remove the uv_random call.  On ziti-router  in tproxy mode it would matter that it was specific to the tproxy port configured for the service.

programs/ziti-edge-tunnel/linux/diverter.c
```diff
@@ -0,0 +1,264 @@
...
+                            if((svc->AllowedSourceAddresses && svc->AllowedSourceAddresses[0] != NULL) || firewall){
+                                diverter_update(svc->Addresses[x]->IP, svc->Addresses[x]->Prefix, svc->Ports[i]->Low, svc->Ports[i]->High, svc->Protocols[j], svc->Id, "-I");
+                            }
+                        }
+                    }
+
+                }
+            }
+        }
+        else if(svc->Permissions.Bind){
```
ekoby, line 163: should this be `else`? Both permissions could be set

ekoby (approved):

PR #1056, opened 2024-12-04
CI test - supply ZET binary instead of building it every time

qrkourier:
In CI this change uses the binary built by cmake.yml. 

Running locally, you must set a var, e.g., `ZITI_EDGE_TUNNEL_BIN=./build/programs/ziti-edge-tunnel/Release/ziti-edge-tunnel ./docker/docker.test.bash`.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,13 +1,6 @@
 name: CI build
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
-      - release-*
   workflow_dispatch:
   workflow_call:
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Test Deployments
+name: Main Test Workflow
 on:
   workflow_dispatch:
   push:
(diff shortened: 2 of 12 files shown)


dovholuknf (approved):

.github/workflows/cmake.yml
```diff
@@ -1,13 +1,6 @@
 name: CI build
 
 on:
-  push:
```
dovholuknf, line 4: actually, was this change intentional?

qrkourier: Yes. This PR changes the way GitHub workflows are triggered. Before, cmake.yml was triggered by push to default and PR HEADs based on main _and_ as a reusable (callable) workflow on release.

Now, cmake.yml is only called by other workflows, and test-deployments.yml takes over calling cmake.yml on push to default and PR HEADs based on main.

qrkourier:
I expect the checks on this branch to succeed after the branch merges on which this twig is based.

.github/workflows/cmake.yml
dovholuknf: Assuming I understand you, this feels inverted to me. i would think the flow is:

- build the artifacts
- upload the artifacts to the job
- kick off other 'test' jobs that download the artifact they want to test/verifiy
- mark the job successful/failed

having something like "test-deployments" initiate the build -- seems wrong. Maybe I'm not understanding you?

qrkourier: Correct. That is the flow. The Test deployments workflow runs on push to default and PR HEADs based on default. We can name it whatever.

PR #1067, opened 2024-12-13
fix config dir handling. allow for relative paths, back pathing etc

dovholuknf:


programs/ziti-edge-tunnel/config-utils.c
```diff
@@ -138,3 +138,28 @@ void update_identity_config(uv_loop_t *l, const char *identifier, const char *cf
...
+char* resolve_directory(const char* path) {
+    char* resolved_path = (char*)malloc(PATH_MAX);
+    if (access(path, F_OK) != -1) {
+        //means the file exists right where it is, use realpath and normalize it and continue
+        if (realpath(path, resolved_path) == NULL) {
+            //how could we get here?
+            printf("path does not exist or permission denied: %s\n", resolved_path);
+            exit(1);
+        }
+    } else {
```
ekoby, line 150: Is this needed? Resolving it won't fix access

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -1324,7 +1324,7 @@ static void run(int argc, char *argv[]) {
 
     // generate tunnel status instance and save active state and start time
     if (config_dir != NULL) {
-        if (!realpath(config_dir, config_dir)) {
+        if (!realpath(config_dir, NULL)) {
```
ekoby, line 1327: Why not resolve_directory?
And it is a memory leak

programs/ziti-edge-tunnel/config-utils.c
dovholuknf: nope. thanks for noticing. will remove

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
dovholuknf: it's now done in option processing so i removed it and the if below

programs/ziti-edge-tunnel/config-utils.c
```diff
@@ -138,3 +148,28 @@ void update_identity_config(uv_loop_t *l, const char *identifier, const char *cf
     }
 }
 
+char* resolve_directory(const char* path) {
```
scareything, line 151: this can/should be `static` since it's not referenced outside of this module. also whitespace.

scareything (approved):

PR #1071, opened 2024-12-20
normalize copyright comments

scareything:


dovholuknf (approved):
not QUITE normalized as there's two distinct formats but much better :)

programs/ziti-edge-tunnel/windows-service.c
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019-2020 NetFoundry, Inc.
+Copyright NetFoundry, Inc.
```
dovholuknf, line 2: oops! check this one

PR #1088, opened 2025-01-17
allow ZITI_DEBUG to be defined

dovholuknf:
if ZITI_DEBUG is declared, define a symbol to trigger debuggier logging in the c sdk

depends on https://github.com/openziti/ziti-sdk-c/pull/811/files

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ else()
     add_compile_definitions(PATH_SEP='/')
 endif()
 
-if(DEFINED ENV{ZITI_DEBUG})
+if(ZITI_DEBUG)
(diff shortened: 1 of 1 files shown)


ekoby:
you don't need this change, you can add compiler flags directly in your cmake profile

CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ else()
     add_compile_definitions(PATH_SEP='/')
 endif()
 
+if(DEFINED ENV{ZITI_DEBUG})
```
scareything, line 21: testing a cmake cache variable instead of the environment would make more sense to me for an option that changes the build.

scareything (approved):

PR #1096, opened 2025-02-05
close server connection when hosted service becomes unavailable

scareything:


lib/ziti-tunnel-cbs/ziti_hosting.c
```diff
@@ -574,6 +574,10 @@ static void on_hosted_client_connect(ziti_connection serv, ziti_connection clt,
     if (status != ZITI_OK) {
         ZITI_LOG(ERROR, "hosted_service[%s] incoming connection failed: %s", service_ctx->service_name, ziti_errorstr(status));
         ziti_close(clt, NULL);
```
ekoby, line 576: when `status != ZITI_OK` - `clt` should be NULL. you can remove this line

Issue #1114, opened 2025-03-17
`listen_options.max_connections` are not passed to SDK


ekoby:
`max_connections` should be copied in [`listen_opts_from_host_cfg_v1`](https://github.com/openziti/ziti-tunnel-sdk-c/blob/main/lib/ziti-tunnel-cbs/ziti_hosting.c#L772)

PR #1115, opened 2025-03-17
set max connection timeout value in listen options

scareything:
also use defaults for int values that are set (to 0) in the config struct even when not set in json


ekoby (approved):
fwiw, connect timeout makes little sense as a listen option. We may ditch at some point

ekoby:
Link #1114?

PR #1118, opened 2025-03-20
coalesce packet buffers before writing to ziti

scareything:
use `pbuf_coalesce` to combine multiple pbufs.

this approach is simpler than writing each chained pbuf separately as the udp code was doing. the previous udp implementation that iterated through the pbufs could lead to double-freeing the chained pbufs.


lib/ziti-tunnel/tunnel_udp.c
```diff
@@ -101,7 +101,12 @@ void on_udp_client_data(void *io_context, struct udp_pcb *pcb, struct pbuf *p, c
 }
 
 void tunneler_udp_ack(struct write_ctx_s *write_ctx) {
-    pbuf_free(write_ctx->pbuf);
+    struct pbuf *p = write_ctx->pbuf;
```
ekoby, line 104: I think it would be more clear if de-chaining happened during writing

ekoby (approved):

PR #1125, opened 2025-04-01
update vcpkg@2025.01.13

scareything:
Newer baselines introduce build failures on our current toolchains.

Try re-running bootstrap-vcpkg.sh/cmd (in your vcpkg checkout) if vcpkg install fails because "vcpkgTools.xml" isn't found.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -28,5 +28,5 @@
       ]
     }
   },
-  "builtin-baseline": "d5ec528843d29e3a52d745a64b469f810b2cedbf"
+  "builtin-baseline": "6f29f12e82a8293156836ad81cc9bf5af41fe836"
 }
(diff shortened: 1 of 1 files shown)


qrkourier:
ziti-builder:v2 _should_ have vcpkg baseline 2025.03.19 as of Mar 23rd when this was released: https://github.com/openziti/ziti-builder/releases/tag/v2.0.12

qrkourier (approved):
There may still be some confusion about the `openziti/ziti-builder:v2` image that is used to build GitHub binaries: it is providing vcpkg, but the composite action installs a redundant copy of vcpk in the runner image where it's already installed ([link to composite action](https://github.com/openziti/ziti-tunnel-sdk-c/blob/main/.github/actions/build/action.yml#L37)), so it may be unexpected there are two redundant, possibly conflicting, instances of vcpkg.

PR #1126, opened 2025-04-01
refine support bundler

qrkourier:


scripts/ziti-edge-tunnel-debug.bash
```diff
@@ -43,6 +44,10 @@ main() {
     # get the aggregate service unit definition
     # shellcheck disable=SC2094
     systemctl cat "${SYSTEMD_SERVICE_UNIT}" > "${SYSTEMD_SERVICE_UNIT}"
+    if [[ -s "${SYSTEMD_ENV_FILE}" ]]; then
+        cp "${SYSTEMD_ENV_FILE}" "$(basename "${SYSTEMD_ENV_FILE}")"
```
scareything, line 48: why use `basename` here? I'd think "." would be equivalent?

scareything (approved):

scripts/ziti-edge-tunnel-debug.bash
qrkourier: Good catch

PR #1136, opened 2025-04-08
disable duplocale when building for darwin

scareything:
the build files in ziti-sdk-swift do this too


dovholuknf (approved):

vcpkg-overlays/omit-json-c-apps/json-c/disable-duplocale.patch
ekoby, line 1: I am not sure about the `vcpkg-overlays` structure. Where is this overlay used?

scareything: We refer to in in CMakePresets.son, in the `vcpkg` preset.

Issue #1159, opened 2025-04-29
preserve renewals in all supported distributions of ZET


qrkourier:
It is becoming increasingly crucial for Edge SDK implementations, such as ZET, to reliably preserve renewed or extended client certificates at runtime. This allows them, for example, to respond to a signal from the controller indicating that their certificate needs to be renewed, or to request certificate renewal when it is approaching expiration.

The Linux distribution of ZET [was recently changed to support this](https://github.com/openziti/ziti-tunnel-sdk-c/pull/1153) by setting a writeable filemode on the `--identity-dir` from which it loads Identity configurations as JSON files.

Next, the Docker and Kubernetes distributions must recommend a writeable, persistent file system as the default storage medium for the enrolled identity and warn when it is not, if possible.

There are differing implications for each method of supplying an identity to ZET in Docker and Kubernetes.

### How Docker ZET Works

- `ZITI_IDENTITY_JSON` - When this is set, the value is written to the identity dir at startup, clobbering any existing file. The file is writeable by ZET, so renewal succeeds and is not preserved. We could deprecate this method or change it so that it does not clobber the existing file.
- `ZITI_ENROLL_TOKEN` - When the identity does not exist and this is set, the value is used to enroll, and the JSON is written to the identity dir. If the identity dir is a persisted volume, then renewals, too, are preserved.
- `ZITI_IDENTITY_BASENAME` - When set, the file `${ZITI_IDENTITY_DIR}/${ZITI_IDENTITY_BASENAME}.json` is assumed to be writeable in a persisted, mounted volume.

### How Kubernetes ZET Works

- `zitiIdentity` - When set, a Secret is created with the JSON and mounted read-only on the container
- `secret.existingSecretName` - When set, no Secret is created and instead the existing Secret is mounted read-only on the container
- `additionalVolumes` - When set, additional Secret(s) and ConfigMap(s) are mounted read-only on the container

### Proposed Changes

- [x] doc - advertise the requirement for a persisted, writeable filesystem
- [x] ~~doc - describe orchestrated alternatives such as an out-of-band re-enrollment procedure when secrets are persisted in a cloud vault~~
- [x] Docker - change entrypoint so `ZITI_IDENTITY_JSON` does not clobber an existing file, allowing renewals to be preserved by mounting a volume on `ZITI_IDENTITY_DIR`
[...]

PR #1160, opened 2025-04-29
don't clobber existing identity with ZITI_IDENTITY_JSON

qrkourier:
relates to https://github.com/openziti/ziti-tunnel-sdk-c/issues/1159

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -60,6 +60,13 @@ fi
 # assign default identity dir if not set in parent env; this is a writeable path within the container image
 : "${ZITI_IDENTITY_DIR:="/ziti-edge-tunnel"}"
 
+if [ "$(stat -c %d:%i "${ZITI_IDENTITY_DIR}")" = "$(stat -c %d:%i "$(dirname "${ZITI_IDENTITY_DIR}")")" ]; then
+    echo "WARN: ${ZITI_IDENTITY_DIR} is not a mounted volume; identity renewals and updates will NOT be preserved across container restarts" >&2
+fi
+if ! [ -w "${ZITI_IDENTITY_DIR}" ]; then
+    echo "WARN: ${ZITI_IDENTITY_DIR} is not writable; identity renewals and updates will NOT be preserved across container restarts" >&2
+fi
+
 # if enrolled identity JSON is provided then write it to a file in the identities dir
 if [[ -n "${ZITI_IDENTITY_JSON:-}" ]]; then
     if [[ -z "${ZITI_IDENTITY_BASENAME:-}" ]]; then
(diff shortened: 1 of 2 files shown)


docker/docker-entrypoint.sh
```diff
@@ -60,16 +60,25 @@ fi
 # assign default identity dir if not set in parent env; this is a writeable path within the container image
 : "${ZITI_IDENTITY_DIR:="/ziti-edge-tunnel"}"
 
+# approximates the functionality of the mountpoint command, which could be installed with util-linux at the expense of 20MB additional image footprint
+if [ "$(stat -c %d:%i "${ZITI_IDENTITY_DIR}")" = "$(stat -c %d:%i "$(dirname "${ZITI_IDENTITY_DIR}")")" ]; then
```
dovholuknf, line 64: this doesn't feel like the place to put this. ZET should be doing this check and printing the warning

docker/README.md
```diff
@@ -1,5 +1,9 @@
 # Run The OpenZiti Tunneler with Docker
 
+:::important
```
dovholuknf, line 3: this is "docusaurus" syntax.  a while back I added processing to docusaurus to handle "github flavored syntax"

Change this to:
```

>[!NOTE]
> Persistent, writeable volumes are ....
```

and it'll render as a note in docusaurus and properly in github

dovholuknf: or WARNING or CAUTION

image

image

docker/README.md
```diff
@@ -24,35 +28,54 @@ volumes:
 
 ### Use an Enrolled Identity from the Environment
 
-You may source an existing identity from the environment.
+You may source an existing identity from an environment variable. The value will be written to the mounted volume and the variable ignored thereafter. The identity file in the mounted volume will be overwritten when the certificate is automatically renewed.
```
dovholuknf, line 31: I started a thoughtful comment about rewriting this. As I was about 80% through it occurs to me that given the new need for a writable volume, all this feels like it should be changed to "put the identity file into the writable volume"

docker/docker-entrypoint.sh
qrkourier: That wouldn't work in this case because there's no way to know it's running in a container which is the only time a mountpoint is applicable.

docker/README.md
qrkourier: We can deprecate the env var method before we decommission it, but I wouldn't support that since the utility and viability of the env var method hasn't changed. The alternative is substantially more complex and error prone.

docker/docker-entrypoint.sh
dovholuknf: I don't think we're on the same page. if the ZET does it, it's irrelevant where it runs. The directories/files either would be, or would not be writable. I don't understand how running in a container matters

docker/README.md
dovholuknf: we can simply remove the doc as a first step. I think we should just remove it entirely. We can discuss the substantial more complexity in person, I don't see it...

PR #1166, opened 2025-05-12
add Linux fips build with dynamic libssl

qrkourier:


vcpkg-overlays/dynamic-libssl3/portfile.cmake
```diff
@@ -0,0 +1 @@
+set(VCPKG_POLICY_EMPTY_PACKAGE enabled)
```
scareything, line 1: I think the overlay should instead be defined in/used by the cpack builds (for distros that provide openssl 3) under `.github/actions/openziti-tunnel-build-action/(redhat|ubuntu*)`. This way the grab-and-go binaries continue to be uniquely identified by the target architecture and nothing else.

Yes, this would mean that you need to use the repos to get a fips-ready binary, but I think that's fair. Keep in mind that we're still (trying to) support ubuntu 18 with the gh-release binaries, and ubuntu 18 does not provide libssl3.

qrkourier: You're right. It makes more sense to take this opportunity to switch the Jammy CPack builder from static to dynamic, since we're eliminating the dependency on the newer version of libssl3 that's unavailable from Jammy's repos.

qrkourier: I plan to mark this ready for review after incorporating Eugene's patch to eliminate the dependency on newer libssl3 and verifying the DEB produced by the updated Jammy builder has the expected libssl3 links.

Meanwhile, I welcome further input on the draft PR.

.github/actions/openziti-tunnel-build-action/ubuntu-22.04/Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN apt-get -y install \
         doxygen \
         git \
         graphviz \
+        libssl-dev \
```
scareything, line 30: rocky 9 also provides openssl 3, with package `openssl-devel`

docs/running-with-FIPS.md
```diff
@@ -1,50 +1,68 @@
...
+>releases that are FIPS certified through Ubuntu Pro.
+
 This document describes how to build and run ziti-edge-tunnel in FIPS mode.
 
 Using FIPS mode requires the following:
-- dynamic linking against OpenSSL library
-- OpenSSL FIPS module
-- OpenSSL FIPS configuration
 
-## Dynamically linking against OpenSSL library
```
scareything, line 11: This could use some restructuring. At the very least to mention that the zet binary from our packages for distros x, y, and z are already linked with dynamic openssl libraries, and all of the building stuff that follows is only needed if you a) aren't on distro x, y, or z, or b) don't want to use our packages for whatever reason.

scripts/ziti-builder.sh
```diff
@@ -62,7 +62,7 @@ function set_workspace(){
     else
         echo -e "INFO: project not mounted on ${WORKSPACE}"\
                 "\nINFO: mounting on ziti-builder container"
-        docker pull "openziti/ziti-builder:${ZITI_BUILDER_TAG:-latest}"
+        docker pull "openziti/ziti-builder:${ZITI_BUILDER_TAG:-latest}" &>/dev/null || true
```
scareything, line 65: Why is it ok for the pull to fail?

.github/actions/openziti-tunnel-build-action/ubuntu-22.04/Dockerfile
qrkourier: Cool, yes. I was contemplating doing the same for RH9, too. I'll see if I can get that one to build and run and validate the expected links.

docs/running-with-FIPS.md
qrkourier: It's in a NOTE admonition at the top, but perhaps the placement could be improved.

scripts/ziti-builder.sh
qrkourier: It's a local cross-build script that supports development of ZET binary and packages and emulates the CI environment. It may also be useful for producing one-offs for a proof of concept.

The pull will fail if the selected builder image exists locally but not in the registry, and the script will raise the exception if the `docker run` command, too, fails to match an image.

scareything (approved):

PR #1194, opened 2025-08-04
lower-case arch in bundle name

scareything:


CMakeLists.txt
```diff
@@ -148,7 +148,7 @@ if(CMAKE_SYSTEM_PROCESSOR STREQUAL AMD64)
 elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL aarch64)
     set(bundle_processor "arm64")
 else()
-    set(bundle_processor ${CMAKE_SYSTEM_PROCESSOR})
+    string(TOLOWER ${CMAKE_SYSTEM_PROCESSOR} bundle_processor)
```
qrkourier, line 151: It could cause us to change occurrences of `ziti-edge-tunnel-Linux_` in `*.zip` filenames appearing in CI to lower `l`, i.e., `ziti-edge-tunnel-linux_`.

scareything: I don't think so. This is just changing the processor part of the bundle name. The main reason to make windows/arm64 consistent with Linux and Darwin.

qrkourier (approved):

PR #1197, opened 2025-08-11
let polkitd satisfy the package dependency for policy kit

qrkourier:


scareything:
does the policykit_version logic in deb/postinst.sh need to be tuned to do the right thing (i.e. nothing) when polkitd is installed instead of policykit?

qrkourier:
> does the policykit_version logic in deb/postinst.sh need to be tuned to do the right thing (i.e. nothing) when polkitd is installed instead of policykit?

No, because we always install the *.rules file, which is ignored by Policy Kit <0.106, so we can continue to plan to delete the *.pkla support from postinstall scriptlet whenever we sunset legacy Policy Kit.

scareything (approved):

PR #1273, opened 2026-02-23
Add Nix and NixOS support

kiriwalawren:
# Nix Support for ziti-edge-tunnel

## Prerequisites

- [Determinate Systems installer](https://zero-to-nix.com/start/install):

This enables flakes and the unified `nix` CLI out of the box. If you already
have Nix installed, make sure `experimental-features = [ "nix-command" "flakes" ];` is in
your `~/.config/nix/nix.conf`.

### Build

```bash
nix build
```

The binary is at `./result/bin/ziti-edge-tunnel`.

### Run without installing

```bash
nix run . -- version
nix run . -- enroll --jwt /path/to/token.jwt --identity /path/to/identity.json
```

## Installing

Add the flake to your NixOS configuration inputs:

```nix
{
  inputs.ziti-edge-tunnel.url = "github:openziti/ziti-tunnel-sdk-c";
}
```

Then enable the module:

```nix
{ inputs, ... }:
{
  imports = [ inputs.ziti-tunnel.nixosModules.default ];

  services.ziti-edge-tunnel = {
    enable = true;

    # Optional: enroll identities at boot
    enrollment.identities = {
      mynetwork = {
        jwtFile = "/path/to/enrollment.jwt";
        # identityFile defaults to /opt/openziti/etc/identities/mynetwork.json
      };
    };
  };
}
```

Installing will set up a systemd service for you.

## Updating dependencies

When a new version of ziti-edge-tunnel is released, update the dependencies:

```bash
# Update the flake.lock file
nix flake update

# Update to the latest release
./nix/update.sh

# Or pin to a specific version
./nix/update.sh v1.11.0
```

Then verify the build:

```bash
nix build
./result/bin/ziti-edge-tunnel version
```

> **Note:** `lwip` and `lwip-contrib` are pinned to stable release branches and
> rarely change. If they do, update their `rev` and `hash` fields in
> `nix/packages/ziti-edge-tunnel.nix` manually.


kiriwalawren:
> Hi and thanks for contributing to the openziti project! I'm not very keen on Nix so please give me some time to absorb this and come up with technical questions.
> 
> In the meantime there are a couple of administrative items that will need to be done before your PR can be merged.
> 
>     1. We require all commits in openziti projects to be signed by the author. Please see https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits.
>        Once you set up signing on your local repo you can rebase against latest and that will force the commits to be signed.
> 
>     2. Could you please "sign" our [contributor license agreement](https://github.com/netfoundry/cla) (by adding a comment to this PR)? edit: please add your CLA agreement comment to this PR _after_ you've rebased.
> 
> 
> Thanks again!

I'll start work on signing my commits. However, I don't really want to sign a CLA unless the OpenZiti team is confident that this is something they want in their repository.

scareything:
Thanks for rebasing and jumping through the cla and all of that!

We do want this! I actually think it would be super to have the nixos build(s) available in the GitHub releases for zet. Do you want to add that to this PR or make a new one for the build workflow?

Thanks again!

kiriwalawren:
> We do want this! I actually think it would be super to have the nixos build(s) available in the GitHub releases for zet. Do you want to add that to this PR or make a new one for the build workflow?

So, something to note about Nix, you don't distribute build artifacts like you do with other distributions. You distribute the "build instructions", then Nix rebuilds the artifact from source on the target host in a deterministic way. The `flake.nix` is how this is accomplished. Users add your flake to their inputs using a GitHub reference. The Flake will be pinned to a specific SHA. Then nix downloads the repository and builds the binary from source. Flakes are a lock file mechanism for Nix inputs.

The user would add the following to their flake inputs:
```nix
{
  inputs.ziti-edge-tunnel.url = "github:openziti/ziti-tunnel-sdk-c";
}
```

There are ways around this, of course, by caching your own builds using a binary cache ([Cachix](https://www.cachix.org/) has a free tier that would work for this), then the user would add your cache to their system. However, I don't really think building the zet binary takes long enough to need it. (I'm assuming "zet" means "ziti-edge-tunnel" here)

That said, I _do_ think we should add build step to your CI and add it as a new required check to verify that the build instructions are working. And that is something I think should happen in this PR. I would be happy to add it.

kiriwalawren:
> However, I don't really think building the zet binary takes long enough to need it.

Another reason you might want a cache is to speed up your CI. After the first run of the Nix build, we can look at how long it takes to build and then you can make a decision about whether or not you want a cache configured.

P.S. I have added the workflow.

kiriwalawren:
I should also note that every time the dependencies of the project get updated or a new version gets released, you will have to run `nix flake update` && `nix/update.sh` script. Otherwise, nix will just keep building the same version over and over again. I couldn't think of a way to automate this without it feeling like "black magic wizardry" and obfuscating what is happening. It is also not a very "nix-y" thing to do.

Edit: Most people automate this with a weekly cron job that creates a new PR.

scareything (approved):

kiriwalawren:
This is great! I'm so excited!

PR #1321, opened 2026-04-23
Add enroll to ipc

dovholuknf:
initial set of changes to add enrollTo* to the ipc commands to support 'add' ing an identity from IPC

### DIFF
--- a/programs/ziti-edge-tunnel/process_cmd.c
+++ b/programs/ziti-edge-tunnel/process_cmd.c
@@ -38,6 +38,7 @@ void free_add_id_req(struct add_identity_request_s * req) {
         free(req->url);
         free(req->key);
         free(req->certificate);
+        free(req->provider);
         free(req);
     }
 }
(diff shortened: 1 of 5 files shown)


programs/ziti-edge-tunnel/process_cmd.c
```diff
@@ -445,13 +733,22 @@ bool process_tunnel_commands(const tunnel_command *tnl_cmd, command_cb cb, void
...
 
-            FILE *outfile;
-            if ((outfile = fopen(new_identifier_path, "wb")) == NULL) {
-                ZITI_LOG(ERROR, "failed to open file %s: %s(%d)", new_identifier_path, strerror(errno), errno);
-                result.error = "invalid file name";
-                result.success = false;
-                free_tunnel_add_identity(&tunnel_add_identity_cmd);
-                break;
+            // For cert/token modes the file must NOT be created on disk until OIDC
+            // succeeds — a 0-byte file at handler time would make the identity
```
scareything, line 737: Would the UI (zdew) notice an empty file? I thought it was looking at the `tunnel_identity` list in `tunnel_status` responses?

dovholuknf: yes it actually causes very confusing problems in the UI. It ends up showing up as an identity that doesn't connect, can't authenticate etc. If the OIDC flow failed a 0-byte file gets written and stays "open" (not 'close'-ed) making it so the zet can't even delete the zero-length file... i was a mess... 

when adding an identity with enrollToToken, the identity file shouldn't be written until the pkce flow completes and the enroll to the controller is successful

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -2823,6 +2825,21 @@ static int add_identity_opts(int argc, char *argv[]) {
             case 'u':
                 tunnel_add_identity_opt->controllerURL = optarg;
                 break;
+            case 'm':
+                if (strcmp(optarg, "none") == 0) {
```
scareything: could use model_enum features instead of strings here. e.g:

```
tnl_enroll_mode m = tnl_enroll_modes.value_of(tunnel_add_identity_opt->enrollMode);
if (m ==  tnl_enroll_mode_Unknown) {
    fprintf(stderr, "unknown --enroll-to-mode value: %s (expected none, cert, or token)\n", optarg);
    errors++;
} else {
    tunnel_add_identity_opt->enrollMode = m;
}
```

I guess you still need the list of valid values for the `fprintf` so either way.

scareything (approved):

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
dovholuknf: works for me :)

PR #1324, opened 2026-04-29
Nix: update flake dependencies and tunnel version

kiriwalawren:


dovholuknf:
thanks for keeping this alive. should we look to update this whenever we make a latest release (aka we mark the release as 'latest' in github)? We should probably have some kind of reminder action that would fail or -- something -- along those lines. None of use nix as daily drivers so it's a bit new to us

kiriwalawren:
> thanks for keeping this alive. should we look to update this whenever we make a latest release (aka we mark the release as 'latest' in github)? We should probably have some kind of reminder action that would fail or -- something -- along those lines. None of use nix as daily drivers so it's a bit new to us

The way we usually deal with updates in Nix land is by creating a weekly cron action that creates a new PR. Which would work for this project as well. But in this instance, something that might be better, would be an action that triggers _when_ a new release is created that opens a new PR. We could get away with this because the the nix dependencies don't drive development in this project. So, we only care about those inputs at release time.

Does the team have an idea of a preference for how they would like to see this problem solved?

scareything:
Yeah, thanks for staying on top of the nixOS support here, @kiriwalawren! I was planning on updating the nix deps soon once we're ready to make a stable/latest ziti-edge-tunnel release.

Personally I'd prefer creating a PR that's triggered when zet is released. I'm guessing we'd only want to merge these PRs for zet releases that are marked as latest. You may have noticed that we do a fair number of prereleases around here before updating `latest`.

But I'm just assuming that the nixOS "releases" are for folks who don't necessarily want the prereleases which can sometimes be a little spicy.

kiriwalawren:
> Yeah, thanks for staying on top of the nixOS support here, @kiriwalawren! I was planning on updating the nix deps soon once we're ready to make a stable/latest ziti-edge-tunnel release.

You're welcome!

> I'm guessing we'd only want to merge these PRs for zet releases that are marked as latest.

I'll create a separate PR for this.

scareything (approved):

PR #1333, opened 2026-05-06
Expand IPC integration tests: MFA flows + Enrollment

CLBRITTON2:
### Tests added
- **MFA**: enable, verify, re-auth after toggling an identity off/on, and remove. Both TOTP and recovery codes are covered, plus invalid-code rejection on each.
- **AddIdentity**: empty JWT, JWT for an identity deleted from the controller, bad filenames (slashes, `..`, over the OS limit).

### Util additions
- `events.WaitFor(t, ctx, op, action, fingerprint)`: drains the event pipe until an event matches and skips the rest. Pulled out of an inline event-drain loop that was duplicated across MFA and identity tests.
- `Overlay.DeleteIdentity`: wraps `ziti edge delete identity`


dovholuknf (approved):
good enough for now but the copy/pasta should be cleaned up in the next pr

PR #1336, opened 2026-05-08
add cmake option to exclude libpcap usage from build

scareything:


programs/ziti-edge-tunnel/CMakeLists.txt
```diff
@@ -3,20 +3,29 @@ project(ziti-edge-tunnel)
 if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
     set(NETIF_DRIVER_SOURCE netif_driver/darwin/utun.c)
 endif()
+option(DISABLE_LIBPCAP_FEATURE "libpcap integration toggle" OFF)
```
ekoby: I don't like negative flags due to mental gymnastics I have to do:
 `-DDISABLE_LIBCAP_FEATURE=ON` -- Ugh

same applies to defines in the code.

I think `-DLIBCAP_FEATURE=[ON/OFF]` reads much better

scareything: I don't disagree. I did it like this for consistency with the libsystemd feature, which started with the idea that the build includes all features by default.

dovholuknf (approved):

PR #1337, opened 2026-05-11
Add integration tests for URL enrollment and external auth

CLBRITTON2:
## URL enrollment
- Adds integration tests for `AddIdentity` via controller URL: happy path, malformed URL, valid non-Ziti endpoint, and duplicate-name rejection
- Reworks `TestMain` to take a persistent `-overlay-home` so the test overlay's PKI survives across runs and can be installed into OS trust.
- When the test overlay's CA is not in OS trust, the URL tests skip with copy-paste install/cleanup commands for Linux, macOS, and Windows.
- Wires CI to seed the PKI via `ziti edge quickstart` and install the CA into OS trust before running the integration suite.

## External auth
- Adds an integration test that drives the controller's built-in OIDC PKCE flow end-to-end against a URL-enrolled identity bound to an `ext-jwt-signer`-backed auth policy.
- Adds a `testutil` OIDC driver (`DiscoverOIDCJWKS`, `DriveControllerOIDC`) that walks `/oidc/authorize` -> `/oidc/login/username` -> `/oidc/authorize/callback` -> ZET's loopback listener with no browser.
- Adds `Overlay` helpers for the ext-jwt-signer / auth-policy / externalId-identity lifecycle, plus `IdentityID` lookup and `PurgeAuthPolicies` / `PurgeExtJwtSigners` to extend the existing "Test"-prefix purge.
- Adds an IPC `GetExternalAuth` helper that parses the `ExternalAuth` response, and a `NeedsExtAuth` field on `IdentityStatus`.

### DIFF
--- a/tests/integration/external_auth_test.go
+++ b/tests/integration/external_auth_test.go
@@ -1,0 +1,96 @@
+/*
+Copyright NetFoundry Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/openziti/ziti-tunnel-sdk-c/tests/integration/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExternalAuth(t *testing.T) {
+	overlay.RequireCATrusted(t)
+	t.Run("onUrlEnrolledIdentityCompletes", testExternalAuthOnUrlEnrolledIdentityCompletes)
+}
+
+func testExternalAuthOnUrlEnrolledIdentityCompletes(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
+	defer cancel()
+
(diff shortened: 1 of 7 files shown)


tests/integration/testutil/overlay.go
```diff
@@ -303,11 +399,73 @@ func (o *Overlay) runZiti(ctx context.Context, args ...string) ([]byte, error) {
...
+
+// PurgeExtJwtSigners deletes every ext-jwt-signer whose name starts with prefix.
+func (o *Overlay) PurgeExtJwtSigners(ctx context.Context, prefix string) error {
+	return o.purgeByPrefix(ctx, "ext-jwt-signers", prefix, o.DeleteExtJwtSigner)
+}
+
+func (o *Overlay) purgeByPrefix(ctx context.Context, entity, prefix string, del func(context.Context, string) error) error {
+	filter := fmt.Sprintf(`name contains "%s"`, prefix)
+	for {
+		out, err := o.runZiti(ctx, "edge", "list", entity, filter, "-j")
```
dovholuknf: why not just "ziti edge delete identities 'name contains $prefix'"? (no prefix fwiw which i assume is why you had all this written)...  also you may need "limit none" on this

tests/integration/external_auth_test.go
```diff
@@ -0,0 +1,96 @@
...
+func testExternalAuthOnUrlEnrolledIdentityCompletes(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
+	defer cancel()
+
+	controllerBase := overlay.ControllerHostPort()
+	signerName := identityNameFor(t) + "-signer"
+	policyName := identityNameFor(t) + "-policy"
+	identityName := identityNameFor(t)
+
+	adminID, err := overlay.IdentityID(ctx, "Default Admin")
```
dovholuknf: let's not use an admin for this... let's make a whole new user

dovholuknf (approved):

PR #1342, opened 2026-05-15
URL enrollment + external auth integration tests

CLBRITTON2:
## Summary
- External auth tests via dex OIDC: single-signer, multi-signer, invalid-provider, no-controller-identity
- URL-based identity enrollment tests
- Tunneler-to-tunneler TCP and UDP data-plane tests using two run-mode ZETs
- Supporting test infrastructure for Linux/Mac/Windows

## New testutil
- `testutil/pkce.go`: launches dex with a generated config + static admin user; waits for OIDC discovery before returning
- `testutil/pkce_flow.go`: `DrivePKCEFlow` walks the authorize → login → approval → callback chain headlessly so tests complete an OIDC handshake without a browser
- `testutil/pause.go`: `Pause(t, message)` interactive breakpoint for stepping through a hung test (Windows)
- `testutil/commands.go`: `AssertValidJwtEnrolledIdentityFile` and `AssertValidUrlEnrolledIdentityFile(mode)` shape-check identity files at every successful enrollment site; URL helper branches on `EnrollMode` so enroll-to-none asserts empty cert/key, enroll-to-cert/token asserts populated
- `scripts/fetch-dex.{sh,ps1}`: clone + `go build` dex from source

## Test infra
- **Elevation pre-check**: `testutil.RequireAdmin()` fails fast on non-elevated Windows / non-root Unix instead of hanging on a pipe ZET can't open
- **Port-bind warning**: probes `localhost:1280` and `:3022` at startup; logs a warning if bound but does not abort - needs more work for controller debugging
- **Parallel ZET start**: zetA and zetB launch concurrently
- **TestMain prefix-purge**: deletes leftover `Test*` identities, auth-policies, and ext-jwt-signers before the suite runs
- **Per-step logging**: every IPC call, overlay command, OIDC step, and event op logs a preamble and a result line with the observed value for debugging
- **ZET startup logging**: `StartZET` logs the exact exec invocation + PID; IPC pipe dial logs attempt 1 and every 20th retry
- **Background-goroutine event buffering**: `EventClient` reads the event pipe continuously into an in-memory slice from dial; `WaitFor` scans the buffer + waits for new events, so tests don't lose events published between dial and the first read
- **Tighter timeouts**: 30s overlay-ready (was 90s), 15s IPC dial (was 60s), 20s `events.WaitFor` cap
- **Verbosity flags**: `-zet-verbosity` (default 4=debug) and `-tlsuv-debug` (default 0=off)
- **PKCE log location**: dex's combined stdout+stderr lands in the same `-zet-log-dir` directory as ZET logs
- **CI PKCE build**: `.github/workflows/main.yml` runs `scripts/fetch-dex.sh` and passes `-pkce-bin` to `go test` on all three runners, so the OIDC tests aren't skipped in CI


tests/integration/external_auth_test.go
```diff
@@ -27,72 +27,250 @@ import (
...
-	events.WaitFor(t, ctx, "identity", "needs_ext_login", identityName)
+	t.Logf("driving PKCE flow (issuer=%s email=%s)", pkce.IssuerURL, pkce.Email)
+	require.NoError(t, testutil.DrivePKCEFlow(ctx, authResp.URL, pkce.IssuerURL, pkce.Email, pkce.Password), "drive PKCE flow")
+	t.Logf("PKCE flow completed; controller should reject because no identity has externalId=%q", pkce.ExternalID)
 
-	status, err := client.GetTunnelStatus(ctx)
-	require.NoError(t, err, "Status after URL AddIdentity\n%s", zet.Logs())
-	entry := status.FindIdentity(identityName)
-	require.NotNil(t, entry, "identity %q not found in Status after URL AddIdentity", identityName)
+	events.WaitFor(t, ctx, "controller", "disconnected", name)
```
dovholuknf, line 135: this is strange. i would NOT expect a 'controller disconnected' event after trying to enroll. I would expect something more like an 'identity' event if anything

tests/integration/external_auth_test.go
```diff
@@ -27,72 +27,250 @@ import (
...
+	name := testutil.IdentityName(t)
+
+	signerName, _ := createPKCESignerAndPolicy(t, ctx, name, pkce.ClientIDs[0])
+	// Deliberately DO NOT create a controller identity with externalId. The
+	// OIDC flow at the IdP still succeeds, but the controller rejects login
+	// since nothing maps test@example.com to a known identity.
+	t.Logf("skipping controller identity creation on purpose (controller should reject because externalId=%q is unmapped)", pkce.ExternalID)
+
+	cleanupCtx := context.Background()
+	t.Cleanup(func() {
```
dovholuknf, line 118: this is  just a style thing - i personally dislike cleaning up like this just because i find this allows the tests to be more brittle than if not dirty. also not cleaning up allows tests to be run over and ovrer against the same controller. That will surface paging issues and all sorts of unexpected side effects

tests/integration/list_identities_test.go
dovholuknf: if the command isn't in the list of commands i don't think it needs to be tested
```
C:\Program Files (x86)\NetFoundry Inc\Ziti Desktop Edge\ziti-edge-tunnel.exe
    enroll               enroll Ziti identity
    run                  run Ziti tunnel (required superuser access)
    run-host             run Ziti tunnel to host services
    on_off_identity      enable/disable the identities information
    enable               enable the identities information
    dump                 dump the identities information
    ip_dump              dump ip stack information
    enable_mfa           Enable MFA function fetches the totp url from the controller
    verify_mfa           Verify the mfa login using the auth code while enabling mfa
    remove_mfa           Removes MFA registration from the controller
    submit_mfa           Submit MFA code to authenticate to the controller
    generate_mfa_codes   Generate MFA codes
    get_mfa_codes        Get MFA codes
    ext-jwt-login        login with ext JWT signer
    tunnel_status        Get Tunnel Status
    refresh              refresh identity
    delete               delete the identities information
    add                  enroll and load the identity
    set_log_level        Set log level of the tunneler
    update_tun_ip        Update tun ip of the tunneler
    update_l2_opts       Enable/disable L2 capability
    service_control      execute service control functions for Ziti tunnel (required superuser access)
    endpoint_sts_change  send endpoint status change message to the tunneler
    version              show version
    help                 this message
```

tests/integration/main_test.go
```diff
@@ -60,52 +69,105 @@ func TestMain(m *testing.M) {
 }
 
 func run(m *testing.M) (int, error) {
-	if err := testutil.EnsureNoExistingZET(); err != nil {
+	if err := testutil.RequireAdmin(); err != nil {
```
dovholuknf, line 72: we should make sure zeta and zetb don't exist still

tests/integration/main_test.go
```diff
@@ -60,52 +69,105 @@ func TestMain(m *testing.M) {
...
 	if err != nil {
 		return 0, fmt.Errorf("create zet temp root: %w", err)
 	}
+	log.Printf("setup: zet temp root %s", zetTempRoot)
 	defer os.RemoveAll(zetTempRoot)
 
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
 
+	log.Printf("setup: starting overlay (zitiBin=%s overlayHome=%s)", zitiBin, overlayHome)
```
dovholuknf, line 89: we should probe to confirm the ports needed are open and if not -- we should possibly pause here and ask the user "are you running ziti on these ports and do you want to continue" if there's a terminal attached. if not then it's in CI (probably) and we should just fail with some error thta the ports are open. this way devs can run ziti in a debugger and still run tese tests?

tests/integration/mfa_test.go
```diff
@@ -69,301 +70,273 @@ func testEnableMFAAcceptsTotpRequiredAuthPolicy(t *testing.T) {
...
-	require.NotNil(t, entry, "identity %q not found in Status after SubmitMFA", name)
-	require.False(t, entry.MfaNeeded, "Status.Identities[%q].MfaNeeded should be false after SubmitMFA", name)
-	require.True(t, entry.MfaEnabled, "Status.Identities[%q].MfaEnabled should remain true after SubmitMFA", name)
-	t.Logf("SubmitMFA ID MfaEnabled=%t MfaNeeded=%t", entry.MfaEnabled, entry.MfaNeeded)
+	submitEvt := events.WaitFor(t, ctx, "mfa", "mfa_auth_status", name)
+	require.True(t, submitEvt.Successful, "mfa:mfa_auth_status Successful=%t after SubmitMFA, want true", submitEvt.Successful)
+	t.Logf("mfa:mfa_auth_status reports Successful=%t after SubmitMFA", submitEvt.Successful)
 }
 
 func testMFAReauthenticationAcceptsRecoveryCode(t *testing.T) {
```
dovholuknf, line 183: have we tested what happens when you exhaust ALL recovery codes too?

tests/integration/pkce_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/openziti/ziti-tunnel-sdk-c/tests/integration/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPKCEUp(t *testing.T) {
```
dovholuknf: not sure this is valuable?

dovholuknf (approved):

PR #1346, opened 2026-05-26
Add enroll-to-cert/token tests and refactor IPC wire types

CLBRITTON2:
## Summary
- Add URL-enrollment integration coverage for enroll-to-cert and enroll-to-token, alongside the existing enroll-to-none flow
- Consolidate every IPC wire type into `ipc_types.go` mirroring ZDEW for the most part, and rework `ipc.go` into a typed command/response client; delete the old `commands.go`
- Restructure external auth tests into two batches (multiple signers, single signer) that each run independently
- Rename the test IdP abstraction from `PKCE` -> `IdP` throughout for clarity
- Gate the enroll-to-cert/token phases on ziti 2.0+ since the `--enroll-to-cert` / `--enroll-to-token` signer flags are 2.0 only

## IPC layer
- `ipc_types.go` (new) holds all wire types: per-command `*Data` payloads, typed `*Function` commands, and typed `*Response` structs, with JSON tags matching the C `TUNNEL_*` macros
- `ipc.go` gains a generic `send[INPUT, RESP]` and a `CommandsClient` with one typed method per command (`AddIdentity`, `ExternalAuth`, `EnableMFA`, `Status`, ...); each returns its typed response
- `EventClient` reads the event pipe on a background goroutine; `WaitForIdentityEvent` / `WaitForControllerEvent` / `WaitForMfaEvent` scan the buffer and block until a matching `Op`/action/fingerprint arrives
- `StatusRaw` returns the daemon's `Data` as raw JSON so the wire-contract test can assert the exact top-level keys of tunnel status

## External auth tests
- `TestExternalAuthMultipleSigners`: working signer plus two decoy extras (each with a unique issuer/JWKS URI so ziti 2.0 validation accepts them); runs the default-policy and named-policy none-flow subtests, then deletes the extras
- `TestExternalAuthSingleSigner`: none subtests, then flips the working signer through enroll-to-cert, enroll-to-token, and both via `UpdateExtJwtSigner`, running the matching subtest after each
- `ExtJwtSignerSpec` carries `EnrollToCert` / `EnrollToToken` bools; `UpdateExtJwtSigner` always sends both flags so a mode can be flipped back off

## Rename
- `PKCE` -> `IdP` across types, helpers, and the `state.idp` field; `DrivePKCEFlow` -> `DriveIdPFlow`
- `testutil/pkce.go` -> `idp.go`, `testutil/pkce_flow.go` -> `idp_flow.go` (git mv, so they show as renames)
- `-pkce-bin` flag -> `-idp-bin`; IdP workdir and log path `pkce/` -> `idp/`

## CI
- `Build PKCE IdP` step -> `Build IdP`; `PKCE_BIN` env -> `IDP_BIN`; `-pkce-bin` -> `-idp-bin`; log upload path `pkce/logs` -> `idp/logs`


tests/integration/testutil/idp.go
```diff
@@ -42,41 +51,41 @@ type PKCE struct {
...
+var DefaultIdPUser = IdPUser{
 	Email:    "test@example.com",
 	Username: "test",
 	UserID:   "08a8684b-db88-4b73-90a9-3cd1661f5466",
 	Password: "password",
 }
 
 // bcrypt of "password" at cost 10.
-const defaultPKCEBcryptHash = `$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W`
+const defaultIdPBcryptHash = `$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W`
```
dovholuknf: is this ever referenced ever again? i only see it used in one other place, I'd just put it in there instead?

tests/integration/testutil/overlay.go
```diff
@@ -563,6 +621,17 @@ func (o *Overlay) PurgeIdentities(prefix string) error {
 	return o.deleteWhere("identities", prefix)
 }
 
+// PurgeIdentitiesByExternalId deletes every identity whose externalId equals the given value.
```
dovholuknf: fwiw "every identity" would be only one identity... It has a uniqueness constrant. :)

tests/integration/testutil/ipc.go
```diff
@@ -206,10 +186,10 @@ func (c *EventClient) readLoop() {
...
 }
 
-// WaitFor blocks until the next event matching op/action/fingerprint arrives
-// and advances past it. Blocks indefinitely; rely on the per-test timeout if
-// the event never comes.
-func (c *EventClient) WaitFor(t *testing.T, op, action, fingerprint string) Event {
+// waitForEvent blocks until the next event matching op/action/fingerprint
+// arrives, advances past it, and returns its raw JSON. Blocks indefinitely;
+// rely on the per-test timeout if the event never comes.
+func (c *EventClient) waitForEvent(t *testing.T, op, action, fingerprint string) json.RawMessage {
```
dovholuknf, line 192: this thing feels like it should have a timeout?

CLBRITTON2: A timeout *can* be added. I tried to avoid putting time limits on individual test pieces and just wrap each individual test in a timeout instead.

dovholuknf (approved):

PR #1347, opened 2026-05-27
Add cross-mode re-enroll rejection tests for external auth

CLBRITTON2:
## Summary
- Add four tests proving a same-name identity can't be re-enrolled under a different mode: none to cert, cert to none, cert to token, token to cert
- Extract the full enroll-to-cert and enroll-to-token flows into reusable `completeEnrollToCert` / `completeEnrollToToken` helpers so the new cross-mode tests share one path with the existing completion tests
- Move the completion-flow helpers to the bottom of the file so the suite reads top-down; no behavior change

## External auth tests
- Each follows the same shape: `UpdateExtJwtSigner` to mode A, full enroll under A, `UpdateExtJwtSigner` to mode B, `AddIdentity` with the same name under mode B, then assert rejection and that the original identity file is still valid for mode A

## Helpers
- `completeEnrollToCert` / `completeEnrollToToken` drive a full flow for a name and return the resulting identifier (working signer must already have that mode enabled)
- `createIdentityData` builds the `AddIdentityData` for a given name and `EnrollMode` against the working signer
- `assertIdentityExistsSameName` asserts the rejection: `Success() == false`, `Code` 500, error contains "identity exists with the same name"

### DIFF
--- a/tests/integration/external_auth_test.go
+++ b/tests/integration/external_auth_test.go
@@ -127,78 +127,6 @@ func TestExternalAuthSingleSigner(t *testing.T) {
 	t.Run("testEnrollToCertThenNoneRejected", c.testEnrollToCertThenNoneRejected)
 }
 
-func (c *extAuthContext) testEnrollToCertCompletes(t *testing.T) {
-	testutil.RunTestWithTimeout(t, func(t *testing.T) {
-		name := testutil.IdentityName(t)
-		t.Cleanup(func() { _ = c.overlay.PurgeIdentityByExternalId(c.idp.ExternalID) })
-
-		c.completeEnrollToCert(t, name)
-	})
-}
-
-// completeEnrollToCert drives a full enroll-to-cert flow for name and returns the
-// resulting identifier. The working signer must already have EnrollToCert enabled.
-func (c *extAuthContext) completeEnrollToCert(t *testing.T, name string) string {
-	controllerBase := c.overlay.ControllerHostPort()
-	mode := testutil.EnrollModeCert
-	identityData := testutil.AddIdentityData{
-		IdentityFilename: name,
-		ControllerURL:    &controllerBase,
-		EnrollMode:       &mode,
-		Provider:         &c.workingSigner.name,
-	}
-	addResp := testutil.AddIdentity(t, c.zet.Commands, identityData)
-	require.True(t, addResp.Success(), "AddIdentity (enroll to cert) should succeed: error=%q\n%s", addResp.Error, c.zet.LogPath())
-	require.NotEmpty(t, addResp.Data.URL, "AddIdentity (enroll to cert) response has empty URL")
-
-	c.idp.DriveIdPFlow(t, addResp.Data.URL)
-
-	added := c.zet.Events.WaitForIdentityEvent(t, "added", name)
-	require.True(t, added.Id.Active, "identity:added Active=%t after enroll to cert IdP login flow, want true", added.Id.Active)
-	require.False(t, added.Id.NeedsExtAuth, "identity:added NeedsExtAuth=%t after enroll to cert IdP login flow, want false", added.Id.NeedsExtAuth)
-	testutil.AssertValidUrlEnrolledIdentityFile(t, added.Id.Identifier, testutil.EnrollModeCert)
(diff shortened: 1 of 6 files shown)


tests/integration/external_auth_test.go
```diff
@@ -315,3 +255,157 @@ func (c *extAuthContext) addEnrollToNoneIdentity(t *testing.T, name string) test
...
+
+		enrollToNone := testutil.ExtJwtSignerSpec{}
+		c.overlay.UpdateExtJwtSigner(t, c.workingSigner.name, enrollToNone)
+		identityEvent := c.addEnrollToNoneIdentity(t, name)
+
+		c.overlay.UpdateExtJwtSigner(t, c.workingSigner.name, testutil.ExtJwtSignerSpec{EnrollToCert: true})
+		enrollToCertIdentity := c.createIdentityData(name, testutil.EnrollModeCert)
+		addResp := testutil.AddIdentity(t, c.zet.Commands, enrollToCertIdentity)
+		c.assertIdentityExistsSameName(t, addResp)
+		testutil.AssertValidUrlEnrolledIdentityFile(t, identityEvent.Id.Identifier, testutil.EnrollModeNone)
```
dovholuknf, line 355: i might have made a "ValidIdentityEnrollToNone" type of func that verifies the identity file doesn't have a key/cert field as well as the url is valid.

tests/integration/external_auth_test.go
```diff
@@ -315,3 +255,157 @@ func (c *extAuthContext) addEnrollToNoneIdentity(t *testing.T, name string) test
...
+		c.overlay.UpdateExtJwtSigner(t, c.workingSigner.name, enrollToNone)
+
+		controllerBase := c.overlay.ControllerHostPort()
+		enrollToNoneIdentity := testutil.AddIdentityData{
+			IdentityFilename: name,
+			ControllerURL:    &controllerBase,
+		}
+		addResp := testutil.AddIdentity(t, c.zet.Commands, enrollToNoneIdentity)
+		c.assertIdentityExistsSameName(t, addResp)
+		testutil.AssertValidUrlEnrolledIdentityFile(t, identifier, testutil.EnrollModeCert)
```
dovholuknf, line 377: also need to verify the key/cert are still here

tests/integration/external_auth_test.go
```diff
@@ -315,3 +255,157 @@ func (c *extAuthContext) addEnrollToNoneIdentity(t *testing.T, name string) test
...
+	})
+}
+
+func (c *extAuthContext) testEnrollToCertThenTokenRejected(t *testing.T) {
+	testutil.RunTestWithTimeout(t, func(t *testing.T) {
+		name := testutil.IdentityName(t)
+		t.Cleanup(func() { _ = c.overlay.PurgeIdentityByExternalId(c.idp.ExternalID) })
```
dovholuknf: `c.overlay.PurgeIdentityByExternalId` could just be an empty arg func and all these become:

```
t.Cleanup(c.overlay.PurgeIdentityByExternalId)
```

tests/integration/external_auth_test.go
```diff
@@ -315,3 +255,157 @@ func (c *extAuthContext) addEnrollToNoneIdentity(t *testing.T, name string) test
...
+
+		c.overlay.UpdateExtJwtSigner(t, c.workingSigner.name, testutil.ExtJwtSignerSpec{EnrollToCert: true})
+		identifier := c.completeEnrollToCert(t, name)
+
+		c.overlay.UpdateExtJwtSigner(t, c.workingSigner.name, testutil.ExtJwtSignerSpec{EnrollToToken: true})
+		enrollToTokenIdentity := c.createIdentityData(name, testutil.EnrollModeToken)
+		addResp := testutil.AddIdentity(t, c.zet.Commands, enrollToTokenIdentity)
+		c.assertIdentityExistsSameName(t, addResp)
+		testutil.AssertValidUrlEnrolledIdentityFile(t, identifier, testutil.EnrollModeCert)
```
dovholuknf, line 393: and verify key/cert still there and probably that they didn't change

dovholuknf (approved):

tests/integration/external_auth_test.go
dovholuknf: nevermind - chris pointed out that it does indeed do this already :)

PR #1349, opened 2026-05-29
Rework integration test harness: JSON config, external controller + IdP support, scripted CI

CLBRITTON2:
## Summary
- Replace the integration suite's CLI flags with a single JSON config (`-config`), with a checked-in `config.example.json` template
- Let the suite target an existing controller by URL (or keep spinning up a local quickstart overlay), and an external IdP (or the bundled test-harness IdP)
- Make external auth IdP-agnostic so the same tests run against dex, Keycloak, and Auth0
- Script the whole CI run (`run-ci.sh` / `run-ci.ps1`) so a dev can reproduce it locally with one command, and run it across a ziti version matrix

## Test config
- Add `testutil.Config` loaded from JSON; `main_test.go` builds the harness from it instead of flags, and every consuming test is adapted
- `ziti.url` selects an external controller; empty falls back to the local `ziti edge quickstart` overlay
- `idp.*` configures the IdP; `zetA` / `zetB` allow two different `ziti-edge-tunnel` binaries

## External auth / IdP
- Discover `jwks_uri` from the IdP's OIDC config instead of hardcoding `issuer + /keys`
- Parse the login form generically (carry hidden state, detect the username/password fields) so dex, Keycloak, and Auth0 all drive from one path
- `idp.signerName` adopts an existing ext-jwt-signer (adopt-or-fail); only the test-harness IdP auto-creates one
- Wire `audience` and `scopes` from config instead of conflating `audience` with the client id and hardcoding scopes

## CI
- Add `run-ci.sh` / `run-ci.ps1` that download ziti, build the IdP, seed PKI, write the config, and run the suite
- Add a `resolve-ziti-versions` job plus `workflow_dispatch` inputs (ziti / ZET / IdP versions); default matrix is newest + newest v2.0 + newest v1.6
- Make the OS-trust CA install opt-in via `--install-cert` / `-InstallCert` (CI passes it by default) so a normal local run doesn't mutate the trust store
- Ignore all local `config*.json` except the template


tests/integration/README.md
```diff
@@ -0,0 +1,236 @@
...
+- **`ziti-edge-tunnel`** - build it from this repo (see the top-level build docs) or download a release from <https://github.com/openziti/ziti-tunnel-sdk-c/releases>.
+- **`dex`** - dex ships no prebuilt binaries, so build it with the helper script (see [The IdP](#the-idp)). Skip this unless you want to run external-auth tests with a local IdP.
+
+## Choosing a mode
+
+Copy one of these, fill in the binary paths, and save it (any name; `config*.json` is gitignored except the example, so `config.json`, `config-local.json`, etc. all work).
+
+### A. Local quickstart, no IdP (minimal)
+
+The smallest run. The harness stands up a controller and runs everything except the external-auth tests (those skip with no IdP configured). No dex, no cert trust needed.
```
dovholuknf, line 32: zetB should be omited from 'the smallest run' imo. Also do all of these defaults still appear in the code? it'd be super if they did so that many of these params are also optional but if not - it's fine

tests/integration/README.md
```diff
@@ -0,0 +1,236 @@
...
+  .\scripts\fetch-dex.ps1
+  .\scripts\fetch-dex.ps1 -Version v2.45.1 -Dest C:\tools\dex
+  ```
+
+  The script prints the built binary path; put it in `idp.binary`. (It needs Go and git on PATH.)
+- **External IdP** - `useTestHarnessIdP: false` with `issuer` + `signerName` (mode C). The login form is read generically, so dex, Keycloak, and Auth0 all work; the IdP user must be a username/password (database) user, not a social login. The signer matches one claim (its `claimsProperty`) against the identity's `externalId`, so `idp.sub` (or `idp.user.email` when `sub` is empty) must equal that claim's value in the token.
+
+## Trusting the test CA
+
+The external-auth tests verify the controller's TLS cert against the OS trust store; if it isn't trusted, they skip. The rest of the suite doesn't need this, because `ziti-edge-tunnel` trusts the controller via its enrollment bundle.
```
dovholuknf: > The external-auth tests verify the controller's TLS cert against the OS trust store

this is only true for "by url enrollment" tests right? it's only necessary if you want to add identities 'byUrl'? (regardless of the 'mode')

CLBRITTON2: Right

tests/integration/README.md
CLBRITTON2: They're not defaulted in code anymore. 
Could add defaults for things like user/password on Ziti quickstarts + ZET verbosity/tlsuv debugging etc

scareything (approved):

PR #1350, opened 2026-06-01
add zet leak tests

scareything:


tests/integration/url_enrollment_test.go
```diff
@@ -141,7 +144,12 @@ func testUrlEnrollmentWithMalformedUrlFails(t *testing.T) {
 		client := testutil.OpenCommandPipe(t, state.zetClient)
 
 		identityName := testutil.IdentityName(t)
-		badURL := "not-a-url"
+		// "not-a-url" looks like a bare hostname with no TLD; on macOS the resolver
```
dovholuknf: i'm curious why wouldn't we want to fix this instead? shouldn't the tunneler deal with this oddball case?

tests/integration/data_path_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+	return interceptZET, hostZET
+}
+
+// TestLeak_TCPTinyEcho iterates 100 connect/echo/close cycles over a TCP
+// service and asserts no lwIP, ziti, FD, or RSS leaks after settle.
+//
+// Requires root/CAP_NET_ADMIN.
+func TestLeak_TCPTinyEcho(t *testing.T) {
+	const (
```
dovholuknf, line 117: shoudl this also have the IsLsanBinary gate?

dovholuknf: (same comment for other tests)

tests/integration/data_path_test.go
```diff
@@ -0,0 +1,333 @@
...
+	harness := testutil.NewLeakHarness(t,
+		[]*testutil.ZET{interceptZET, hostZET}, tcpLeakPools)
+	payload := []byte("hello-ziti-leak-tcp")
+
+	harness.Run(t, func() {
+		for i := range iterations {
+			conn, err := net.DialTimeout("tcp", interceptAddr, 10*time.Second)
+			require.NoError(t, err, "iter %d: dial\ninterceptZET: %s\nhostZET: %s",
+				i, state.zetClient.LogFile(), state.zetHost.LogFile())
+			_ = conn.SetDeadline(time.Now().Add(10 * time.Second))
```
dovholuknf, line 145: really ignore all these errors? presumably it fails for other reasons but it's not useful to konw which error caused a fail?

tests/integration/data_path_test.go
```diff
@@ -0,0 +1,333 @@
...
+		interceptPort = 24000
+	)
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	interceptZET, hostZET := startLeakZETs(t)
+	echo := testutil.StartUDPEcho(t)
+	names := t2tNames(t)
+	setupT2TService(t, ctx, interceptZET, hostZET, names,
+		"udp", "127.0.0.1", addrPort(t, echo.Addr), interceptIP, interceptPort)
```
dovholuknf, line 228: not sure if you plan to test "localhost" in particular with ipv4/ipv6 bindings but that's caught us before with oddball situations where each os does things differently? just a thought

dovholuknf (approved):

tests/integration/url_enrollment_test.go
scareything: I'm not convinced that this actually fixes anything tbh. I'm currently trying to get the tests to complete on macOS runners without hanging or OOMing the gh runner agent.

PR #1352, opened 2026-06-05
Integration Tests: fixture-seeded identities via ziti ops import, ZET restart coverage, harness-managed CA trust

CLBRITTON2:
## Summary
- Seed all controller-side test identities from a checked-in fixture via `ziti ops import` instead of creating them per test. Follow-up: migrate the remaining controller-side test entities (auth policies, services, service policies, configs, ext-jwt-signers - everything `ziti ops import` supports) into the fixture the same way
- Add ZET restart coverage: enrolled, MFA-verified, ext-auth, and disabled identity state survives a ZET restart
- Give every ext-auth test its own IdP user so controller identities never share an `externalId`
- Let the harness manage OS trust for the overlay CA  via config file `autoTrustCa`, with a fail-fast guard for ziti 1.6's `ops import` trust bug
- Consolidate the IPC test utilities so call sites stay small: helpers log and assert the send internally, callers assert outcomes with chainable `AssertSuccess` / `AssertFail`

## Identity setup
- Add `testdata/fixture.json` imported once during setup; tests fetch the pending enrollment JWT from the controller (`GetJwtFromController`) and enroll it on the ZET
- Tests use literal identity names (`test_mfa_verify_invalid_totp`, ...) instead of deriving them from `t.Name()`, so renaming a test won't orphan fixture entries
- Setup purges stale `test_*` entities and auto-provisioned `@test.com` identities before importing, so reruns start clean

## Restart coverage
- New `tunnel_restart_test.go`: enroll a JWT identity, an ext-auth identity, an MFA-verified identity, and a disabled identity, restart the ZET, and assert each state is restored from the status/events pushed on reconnect - the goal is to have a test that has *many* identities in many *different states* that all maintain an expected state on ZET restart.
- Add `ReconnectEvents` to re-subscribe to the restarted ZET's event pipe

## External auth
- `dex-config.yaml` defines one user per ext-auth test (email = the identity's `externalId`); tests no longer share a login

## CA trust
- `ziti.autoTrustCa: true` (quickstart mode): the harness installs the overlay CA into OS trust during setup and removes it at teardown; the install deliberately happens after the fixture import
- Setup fails fast when a ziti 1.6 CLI meets an OS-trusted controller (its import login builds an empty TLS root pool), and the URL-enrollment tests skip pointing at `autoTrustCa` instead of handing out manual install commands; both checks carry a TODO to remove when 1.6 leaves LTS

## Test utilities
- IPC command methods log the command and assert the send succeeded; failure tests assert the specific code and error reason
- `runtest.go` becomes `timeout.go` (`RunWithTimeout` / `RunWithTimeoutOf`)


tests/integration/testutil/ipc.go
```diff
@@ -120,29 +108,36 @@ func openCommandPipe(path string, done <-chan struct{}) (*CommandsClient, error)
...
+			log.Printf("ipc: dial event pipe connected %s", path)
 			return nil, fmt.Errorf("process exited before event pipe %s became dialable: %v", path, err)
 		case <-time.After(dialRetryInterval):
 		}
 	}
 }
 
+// ReconnectEvents closes the event pipe and re-subscribes to the same running
+// daemon, replacing z.EventClient. The daemon resends its status snapshot on connect.
+func (z *ZET) ReconnectEvents(t *testing.T) {
```
dovholuknf: this seems like the wrong place to put this func since it's using the ZET receiver?

tests/integration/testutil/ipc.go
```diff
@@ -252,176 +247,239 @@ func (c *EventClient) WaitForMfaEvent(t *testing.T, action, fingerprint string)
...
+	return ev
+}
+
 func (c *EventClient) Close() error {
 	return c.conn.Close()
 }
 
 // DialIPC connects to this ZET instance's IPC command pipe.
 // Retries until the pipe is dialable or the ZET process exits.
 func (z *ZET) DialIPC() (*CommandsClient, error) {
```
dovholuknf: another ZET receiver - scan the whole file for that sorta thing

tests/integration/testutil/overlay.go
```diff
@@ -234,10 +235,51 @@ func (o *Overlay) OSCAStrings() (install, cleanup string) {
 	return
 }
 
+// InstallCATrust installs the quickstart CA into the OS trust store.
+func (o *Overlay) InstallCATrust() error {
+	if runtime.GOOS == "windows" {
+		return runCommand("certutil", "-addstore", "-f", "Root", filepath.Join(o.Home, "pki", "root-ca", "certs", "root-ca.cert"))
```
dovholuknf, line 243: this is something that might want more comments about WHY we have this. or not...

tests/integration/testutil/overlay.go
```diff
@@ -234,10 +235,51 @@ func (o *Overlay) OSCAStrings() (install, cleanup string) {
...
+
 // RequireCATrusted skips the test (with OS-specific install/cleanup
 // instructions) if the overlay's CA isn't in the calling OS's trust store.
 func (o *Overlay) RequireCATrusted(t *testing.T) {
-	if o.CATrusted() {
+	// With AutoTrustCA the harness installed the CA during setup
+	if o.AutoTrustCA || o.CATrusted() {
+		return
+	}
+	// TODO: remove this branch when ziti 1.6 is no longer part of LTS; manual
```
dovholuknf, line 280: this todo should be just baked into the test so that when it updates and is no longer < 2 you just get a test failure saying "remove this test and fix that other thing cause it's not needed any more"

tests/integration/testutil/timeout.go
```diff
@@ -41,3 +41,22 @@ func RunTestWithTimeout(t *testing.T, f func(t *testing.T)) {
 		t.Fatalf("test %s timed out after %s", t.Name(), TestTimeout)
 	}
 }
+
+func RunWithTimeoutOf(t *testing.T, timeout time.Duration, f func(t *testing.T)) {
```
dovholuknf, line 31: totally duplicates RunWithTimeout. I would make RunWithDefaultTimeout call this function (and rename the function but i'd prolly not rename the func in this pr... just call RunWithTimeoutOf from RunWithTimeout -- passing the default TestTimeout...

tests/integration/add_identity_test.go
```diff
@@ -25,200 +25,108 @@ import (
...
-		overlay := state.overlay
-		events := state.zetClient.Events
-		client := state.zetClient.Commands
-
-		identityName := testutil.IdentityName(t)
-		t.Logf("creating JWT for %q", identityName)
-		jwt, err := overlay.CreateIdentityJWT(identityName)
-		require.NoError(t, err, "failed to create JWT via overlay")
-		require.NotEmpty(t, jwt, "JWT content should not be empty")
+func sameJwtTwiceSecondFails(t *testing.T) {
```
dovholuknf, line 47: TwiceSecond reads funny... but i suppose i understand. At first it felt like  a typo tho

tests/integration/url_enrollment_test.go
```diff
@@ -20,136 +20,76 @@ import (
 	"testing"
 
 	"github.com/openziti/ziti-tunnel-sdk-c/tests/integration/testutil"
-	"github.com/stretchr/testify/require"
 )
```
dovholuknf: these should be in 'add_identity_test' OR the test should be renamed to somethhing like 'add_identity_by_url_test'

tests/integration/url_enrollment_test.go
```diff
@@ -20,136 +20,76 @@ import (
...
+			IdentityFilename: "test_url_enroll_non_ziti",
 			ControllerURL:    &nonZitiURL,
 		}
 
-		resp := testutil.AddIdentity(t, client, identityData)
-		require.False(t, resp.Success(), "non-Ziti URL %q should be rejected, got Success=true\n%s", nonZitiURL, state.zetClient.LogFile())
-		require.Equal(t, 500, resp.Code, "expected Code=500, got %d", resp.Code)
-		t.Logf("non-Ziti URL rejected: code=%d error=%q", resp.Code, resp.Error)
+		addResp := state.zetClient.AddIdentity(t, identityData)
+		addResp.AssertFail(t, 500, "")
```
dovholuknf: there's nothing else interesting to assert on here? there are no other events on any of these ?

tests/integration/set_log_level_test.go
```diff
@@ -20,21 +20,14 @@ import (
...
-		client := testutil.OpenCommandPipe(t, state.zetClient)
-
-		t.Logf("sending SetLogLevel %q", "trace")
-		resp, err := client.SetLogLevel("trace")
-		require.NoError(t, err, "failed to send SetLogLevel\n%s", state.zetClient.LogFile())
-		require.True(t, resp.Success(), "SetLogLevel failed: error=%q code=%d", resp.Error, resp.Code)
-		t.Logf("SetLogLevel succeeded: code=%d", resp.Code)
+func succeeds(t *testing.T) {
+	testutil.RunWithTimeout(t, func(t *testing.T) {
+		state.zetClient.SetLogLevel(t, "trace").AssertSuccess(t)
```
dovholuknf: in retrospect i think i'd prefer to see that .Assert on a different line. as i was scanning i didn't even notice it.

tests/integration/remove_identity_test.go
```diff
@@ -25,41 +25,18 @@ import (
...
-
-		t.Logf("sending RemoveIdentity for Identifier=%s", event.Id.Identifier)
-		removeResp, err := client.RemoveIdentity(event.Id.Identifier)
-		require.NoError(t, err, "failed to send RemoveIdentity\n%s", state.zetClient.LogPath())
-		require.True(t, removeResp.Success(), "RemoveIdentity failed: error=%q code=%d", removeResp.Error, removeResp.Code)
-		t.Logf("RemoveIdentity succeeded for %q", name)
+		client.RemoveIdentity(t, event.Id.Identifier).AssertSuccess(t)
 
 		_, statErr := os.Stat(event.Id.Identifier)
 		require.True(t, os.IsNotExist(statErr), "identity file should be removed after RemoveIdentity: %s\n%s", event.Id.Identifier, state.zetClient.LogPath())
```
dovholuknf, line 39: there should be an assertion that some sort of identity event happns too?

tests/integration/mfa_test.go
```diff
@@ -32,364 +25,129 @@ import (
...
-func testEnableMFAAcceptsJwtEnrolledIdentity(t *testing.T) {
-	testutil.RunTestWithTimeout(t, func(t *testing.T) {
-		enrolled, _ := newEnrolledMFA(t, testutil.IdentityName(t))
+func acceptsJwtEnrolledIdentity(t *testing.T) {
+	testutil.RunWithTimeout(t, func(t *testing.T) {
+		enrollment, _ := testutil.EnrollAndEnableMFA(t, state.overlay, state.zetClient, "test_mfa_enable_jwt")
 
-		require.False(t, enrolled.IsVerified, "EnableMFA Data.IsVerified should be false before verify_mfa")
-		t.Logf("EnableMFA returned IsVerified=%t", enrolled.IsVerified)
+		require.False(t, enrollment.IsVerified, "EnableMFA Data.IsVerified should be false before verify_mfa")
```
dovholuknf, line 53: i assume EnrollAndEnableMFA tests for things like 'mfaNeeded' and 'mfaEnabled' too?

tests/integration/mfa_test.go
```diff
@@ -32,364 +25,129 @@ import (
...
+func reauthAcceptsValidTotp(t *testing.T) {
+	testutil.RunWithTimeout(t, func(t *testing.T) {
+		idName := "test_mfa_reauth_valid_totp"
+		enrollment, secret := testutil.EnrollAndVerifyMFA(t, state.overlay, state.zetClient, idName)
 
-func testMFAReauthenticationAcceptsValidTotp(t *testing.T) {
-	testutil.RunTestWithTimeout(t, func(t *testing.T) {
-		name := testutil.IdentityName(t)
-		enrolled, events := newEnrolledMFA(t, name)
+		state.zetClient.DisableEnableIdentity(t, enrollment.Identifier)
```
dovholuknf, line 81: ~~this test passes after calling this? is this just 'toggling' the state ?~~

oh i get it -- it disables, THEN enables the identity... just hard to understand with no ide :)

tests/integration/mfa_test.go
```diff
@@ -32,364 +25,129 @@ import (
...
-	counter := uint64(at.Unix() / 30)
-	msg := make([]byte, 8)
-	binary.BigEndian.PutUint64(msg, counter)
-	mac := hmac.New(sha1.New, key)
-	mac.Write(msg)
-	sum := mac.Sum(nil)
-	offset := sum[len(sum)-1] & 0x0f
-	code := binary.BigEndian.Uint32(sum[offset:offset+4]) & 0x7fffffff
-	return fmt.Sprintf("%06d", code%1_000_000), nil
+func removeRejectsInvalidTotp(t *testing.T) {
```
dovholuknf, line 160: i presume it doesn't matter since it is covered by this test, but no "invalid recovery code" test? And i would test submitting the SAME recovery code twice (second time it should fail)

dovholuknf (approved):
overall looks a lot, lot better! some things to fix here/there

tests/integration/remove_identity_test.go
CLBRITTON2: I see no identity event when I remove an identity in ZDEW. Just the command response.

PR #1401, opened 2026-08-11
Add third-party CA enrollment integration tests

CLBRITTON2:
Adds integration tests for enrolling with client certificates signed by a registered 3rd-party CA: 
- Auto CA enrollment (no pre-created identity)
- OTT CA enrollment (a one-time token presented with a client certificate signed by the CA)
- Rejection of a certificate signed by an unregistered CA.

### DIFF
--- a/tests/integration/testutil/overlay.go
+++ b/tests/integration/testutil/overlay.go
@@ -571,17 +571,18 @@ type CAResult struct {
 	Name string
 }
 
-// caPkiPath is the on-disk PKI directory for a CA created by CreateLocalPkiCA.
-func (o *Overlay) caPkiPath(name string) string {
-	return filepath.Join(o.Home, "third-party-pki", name)
+// pkiRoot is the on-disk root under which all test CAs are created; the ziti
+// pki CLI lays out each CA at <root>/<ca-name>.
+func (o *Overlay) pkiRoot() string {
+	return filepath.Join(o.Home, "third-party-pki")
 }
 
 // CreateLocalPkiCA creates a root CA on disk without registering it on the
 // controller.
 func (o *Overlay) CreateLocalPkiCA(t *testing.T, name string) {
 	// ziti pki refuses to overwrite PKI left by a previous run
-	require.NoError(t, os.RemoveAll(o.caPkiPath(name)), "clear stale pki for %s", name)
-	_, err := o.execZiti("pki", "create", "ca", "--pki-root", filepath.Join(o.Home, "third-party-pki"), "--ca-file", name, "--ca-name", name)
+	require.NoError(t, os.RemoveAll(filepath.Join(o.pkiRoot(), name)), "clear stale pki for %s", name)
+	_, err := o.execZiti("pki", "create", "ca", "--pki-root", o.pkiRoot(), "--ca-file", name, "--ca-name", name)
 	require.NoError(t, err, "pki create ca %s", name)
 }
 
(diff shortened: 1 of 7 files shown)


tests/integration/testutil/overlay.go
```diff
@@ -555,26 +567,135 @@ func (o *Overlay) CreateIdentityWithExternalId(t *testing.T, name, externalID, a
...
+// pki CLI lays out each CA at <root>/<ca-name>.
+func (o *Overlay) pkiRoot() string {
+	return filepath.Join(o.Home, "third-party-pki")
+}
+
+// CreateLocalPkiCA creates a root CA on disk without registering it on the
+// controller.
+func (o *Overlay) CreateLocalPkiCA(t *testing.T, name string) {
+	// ziti pki refuses to overwrite PKI left by a previous run
+	require.NoError(t, os.RemoveAll(filepath.Join(o.pkiRoot(), name)), "clear stale pki for %s", name)
```
dovholuknf: don't love that it removeall's here. seems like the 'create' func shouldn't do destructive operations without a boolean at least passed to do so

tests/integration/testutil/overlay.go
```diff
@@ -555,26 +567,135 @@ func (o *Overlay) CreateIdentityWithExternalId(t *testing.T, name, externalID, a
...
+	o.CreateLocalPkiCA(t, name)
+
+	caCert := filepath.Join(o.pkiRoot(), name, "certs", name+".cert")
+	out, err := o.execZiti("edge create ca %s %s --auth --ottca --autoca", name, caCert)
+	require.NoError(t, err, "create ca %s", name)
+	id := string(bytes.TrimSpace(out))
+
+	// --cacert/--cakey makes the CLI fetch the verificationToken and mint the CN=token cert itself
+	caKey := filepath.Join(o.pkiRoot(), name, "keys", name+".key")
+	_, err = o.execZiti("edge verify ca %s --cacert %s --cakey %s", name, caCert, caKey)
```
dovholuknf: i'd like to do this using the other path as well (possibly instead of) this path. meaning using the ziti cli to make a client cert and using `ziti edge verify ca <name> ( --cert <pemCertFile>` instead of cacert/key

tests/integration/testutil/overlay.go
```diff
@@ -555,26 +567,135 @@ func (o *Overlay) CreateIdentityWithExternalId(t *testing.T, name, externalID, a
...
+	require.NoError(t, err, "pki create client %s for ca %s", commonName, caName)
+	cert, err := os.ReadFile(filepath.Join(o.pkiRoot(), caName, "certs", commonName+".cert"))
+	require.NoError(t, err, "read client cert for %s", commonName)
+	key, err := os.ReadFile(filepath.Join(o.pkiRoot(), caName, "keys", commonName+".key"))
+	require.NoError(t, err, "read client key for %s", commonName)
+	return string(cert), string(key)
+}
+
+// GetCaJwt fetches the CA enrollment JWT from the management API, authenticated
+// with the session the CLI cached at login. No ziti CLI verb exposes it.
```
dovholuknf, line 678: > No ziti CLI verb exposes it.

pleaes file an issue to expose it.

tests/integration/third_party_ca_test.go
```diff
@@ -0,0 +1,88 @@
...
+		caID := state.overlay.CreateThirdPartyCA(t, "test_tpca_neg")
+		caJwt := state.overlay.GetCaJwt(t, caID)
+		state.overlay.CreateLocalPkiCA(t, "test_tpca_unregistered")
+		cert, key := state.overlay.CreateClientCert(t, "test_tpca_unregistered", "test_tpca_unregistered_user1")
+
+		identityData := testutil.NewCaIdentityData("test_tpca_unregistered_user1", caJwt, cert, key)
+		addResp := state.zetClient.AddIdentity(t, identityData)
+		addResp.AssertFail(500, "key/cert are invalid")
+	})
+}
```
dovholuknf, line 148: needs moar tests. adding the same identity twice, having no external id that matches - there seem like a lot of negative tests we could run (and maybe more positive tests)

dovholuknf (approved):
i'd like more tests (positive and negative) if we can get em but i'll approve as is

PR #1403, opened 2026-08-13
Fix legacy auth with mfa on 1.18+

### LINKED ISSUE #1404: MFA-enrolled identity is asked to enroll again on legacy auth with tunneler 1.18+
On an overlay using legacy auth, the tunneler tells clients an identity needs to enroll in MFA even if it is already
enrolled. The client cannot enroll again, so users hit errors trying to enable MFA they already have.

Affects any controller that does not advertise OIDC: v1.x by default, and v2.x configured with
`disableOidcAutoBinding`. Reproduced on v1.1.3 and v1.6.19 with ziti-edge-tunnel 1.18.x.

## Cause

Legacy controllers never send `isTotpEnrolled`, and an absent field now parses as `false`. In 1.18+ ziti-sdk-c
branches on that field to choose between prompting for a code and requesting enrollment, so it cannot tell "not
enrolled" from "not stated" and asks every identity to enroll. The tunneler takes that as authoritative and
overwrites its own `MfaEnabled`, which was correct.

## Reproduce

1. Enroll an identity in MFA against a legacy-auth controller and verify it.
2. Restart the tunneler.
3. ZDEW requires the user to "enable MFA" even though they have already enrolled successfully.

## Workaround until fixed

Bind `edge-oidc` on the controller. Clients then take the OIDC path, the controller states enrollment, and the
identity is asked for a code.

dovholuknf:
Fixes #1404

Old controllers have no `isTotpEnrolled` field, so they never say whether an identity has MFA enrolled. New
controllers do. Tunnelers from 1.18 read that field, and a missing one reads as `false`, so against an old
controller every identity looks unenrolled. The sdk asks it to enroll, and the tunneler believed that over its own
record.

It no longer does. An enroll request now means only that MFA is needed. The tunneler keeps its `MfaEnabled` and
writes it only on verification, a successful login, or a removal. If it knows the identity is enrolled it asks for
a code. If it does not know, and the enroll attempt fails to authenticate, only an enrolled identity fails that
way, so it marks the identity enrolled and asks for a code.

Not writing to disk matters as much as not clearing the flag. An MFA event can arrive before the identity's stored
state loads, 31 times out of 88 in one run, and the old code persisted whatever it had then. It wrote `false` for
an enrolled identity, the next restart read it back, and the client never recovered.

## Tests

The tests came first and failed for the right reason, then passed with the fix, against ziti v1.6.19, v2.0.2 and
v2.1.0-pre1.

`legacy_auth_mfa_test.go` has eleven cases: the prompt after a restart and after an identity toggle, stored state
surviving both, recovery when that state is wrong, two restarts in a row, recovery codes, silence once a code is
accepted, two identities not treading on each other, MFA removed by the client and by an administrator, and a
controller that gains OIDC underneath an enrolled identity. Nine run on both auth paths; two need a controller
that withholds enrollment state and skip on OIDC.

`Overlay` now takes an `Auth` of `OIDC` or `Legacy` and checks the controller advertises what was asked for, so a
run cannot quietly test the wrong path. Fixtures load before the switch, because `ziti ops import` panics against
a controller without the OIDC endpoints. The set costs two minutes, so it runs nightly behind the `slowtests` tag,
in a new `main-legacy-auth` topology.

Restarting the overlay mid-run turned up four harness bugs, fixed here: no wait for the router to enroll before
restarting, `Kill` failing on Windows with the process still holding the port, `Signal(0)` proving nothing there,
and the cluster-leader wait skipped for local builds reporting `v0.0.0`.


tests/integration/testutil/ipc.go
```diff
@@ -173,13 +173,15 @@ func (c *EventClient) readLoop() {
 // waitForEvent blocks until the next event matching op/action/fingerprint
```
CLBRITTON2: Looks like the comment for `waitForEvent` is now out of place

dovholuknf: fascinating. good catch :)

tests/integration/testutil/overlay_auth.go
```diff
@@ -0,0 +1,389 @@
...
+// RequireLegacyAuth skips the test unless this overlay is running on the legacy auth path.
+func (o *Overlay) RequireLegacyAuth(t *testing.T) {
+	o.requireAuthApplied(t)
+	if o.Auth != AuthLegacy {
+		t.Skipf("ziti.auth=%s; this test covers the legacy api-session auth path", o.Auth)
+	}
+}
+
+// RequireOidcAuth skips the test unless this overlay is running on the OIDC auth path.
+func (o *Overlay) RequireOidcAuth(t *testing.T) {
```
CLBRITTON2, line 373: Does anyone/anything call this? same with ControllerCapabilities above?
I think enroll to cert/token tests at least need this for legacy auth tests?

tests/integration/testutil/overlay.go
CLBRITTON2: Legacy auth tests calling this function X times will truncate the first/subsequent controller's log which covers the fixture import among other things. Just a note.

CLBRITTON2:
The [integration test README](https://github.com/openziti/ziti-tunnel-sdk-c/blob/main/tests/integration/README.md) might need updating with these changes

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -711,13 +711,28 @@ static void on_event(const base_event *ev) {
                 break;
             }
             mfa_status s = mfa_statuss.value_of(mfa_ev->operation);
-            set_mfa_status(ev->identifier, (s != mfa_status_enrollment_required), true);
+            const char *action = mfa_ev->operation;
```
ekoby, line 722: this logic should be in tunneler library so that downstreams projects (not using ziti-edge-tunnel executable) behave consistently.

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -711,13 +711,28 @@ static void on_event(const base_event *ev) {
...
+
+            // An enrollment request is not evidence that the identity is unenrolled. Controllers on the
+            // legacy api-session path never report totp enrollment, so the sdk asks to enroll either way,
+            // and this event can arrive before the identity's stored state has been read back. This path
+            // records only that a code is needed: it never clears MfaEnabled and never persists.
```
ekoby: > it never clears MfaEnabled

should be cleared if MFA is removed, no?

tests/integration/testutil/overlay_auth.go
dovholuknf: removed thx

tests/integration/testutil/overlay.go
dovholuknf: yep. it appends now

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
dovholuknf: yah that's a good nuance. if it's false, mfaEnabled should end up being forced back to false ideally with some sort of log and surfacing that to the user somehow. I'll have a think on that to see if it impacts the ZDEW UI

dovholuknf: i'll address this comment. it's caused the confusion

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
dovholuknf: i'll put a different pr up for this ...

dovholuknf:
adding the 60s hack / wait to the docker-based tests causes much of the flakiness in these tests to pass it seems... i'll wait for https://github.com/openziti/ziti-sdk-c/pull/1120 and if we can get a new sdk i'll update to that sdk and remove the hack

CLBRITTON2 (approved):

PR #1405, opened 2026-08-14
Fix Windows route API error reporting

tomc-xsec:
This change updates Windows route API error handling to report the error codes returned directly by the Windows IP Helper APIs rather than using `GetLastError()`. The route APIs return `NO_ERROR` on success and an error code directly on failure. Refer to [Microsoft documentation for `CreateIpForwardEntry2`](https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-createipforwardentry2#return-value) for documented return behavior.

The existing behavior can result in misleading messages such as:

```text
WARN ziti-edge-tunnel:tun.c:550 refresh_routes() failed to create exclusion route[1.2.3.4]: 1168(The operation completed successfully.)
```
Here, the route API returned error `1168`, while `GetLastError()` yielded a value corresponding to `The operation completed successfully`.

The change also verifies that `FormatMessage()` returns `len > 0` before using the formatted error message, avoiding use of an empty or unavailable message.

### DIFF
--- a/programs/ziti-edge-tunnel/netif_driver/windows/tun.c
+++ b/programs/ziti-edge-tunnel/netif_driver/windows/tun.c
@@ -511,7 +511,7 @@ static int tun_exclude_rt(netif_handle dev, uv_loop_t *loop, const char *dest) {
     MIB_IPFORWARD_ROW2 *route = calloc(1, sizeof(MIB_IPFORWARD_ROW2));
     route->DestinationPrefix.Prefix.si_family = AF_INET;
     parse_route(&route->DestinationPrefix, dest);
-    int rc = GetIpForwardEntry2(route);
+    DWORD rc = GetIpForwardEntry2(route);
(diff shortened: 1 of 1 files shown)


ekoby (approved):

dovholuknf:
Hi @tomc-xsec, would you kindly reply with a comment accepting the CLA? I think your name may have changed and we don't have your CLA acceptance on file. You just need to reply with:

"I have read the CLA Document and I hereby sign the CLA" and it should clear. Read the CLA here https://github.com/netfoundry/cla/raw/main/v1.1/NetFoundry-ICLA-v1.1.pdf

tomc-xsec:
I have read the CLA Document and I hereby sign the CLA

scareything:
Thanks for your contribution! Please feel free to add yourself to the `CONTRIBUTORS` file in this PR.

PR #1417, opened 2026-08-27
Add outage/retry test

scareything:
Add a slow test that forces the tunnelers to go through a tcp proxy. The proxy allows the test to disrupt and restore the network between the tunneler and the controller/ER. This makes it possible to verify that the tunneler resumes auth/communication with the overlay after a network outage is restored.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -51,12 +51,6 @@ jobs:
             preset: linux-arm64
 
     steps:
-      - name: Debug action
-        shell: bash
-        run: |
-          env | sort
-          cat "$GITHUB_EVENT_PATH"
-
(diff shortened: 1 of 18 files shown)


scareything:
the nightly tests:
- pass with ziti-sdk-c@1.18.7: https://github.com/openziti/ziti-tunnel-sdk-c/actions/runs/33118182183
- fail with ziti-sdk-c@1.18.5:https://github.com/openziti/ziti-tunnel-sdk-c/actions/runs/33120067806
- pass again when ziti-sdk-c rollback is undone: https://github.com/openziti/ziti-tunnel-sdk-c/actions/runs/33122175070

dovholuknf (approved):

PR #1437, opened 2026-09-18
Reject update_tun_ip with a missing IP

### LINKED ISSUE #1436: update_tun_ip without a tun IP crashes the tunneler
Sending the update_tun_ip IPC command with a prefix length but no tun IP crashes ziti-edge-tunnel. A malformed
IPC command should come back as an error, not take the service down. 

Reproduce on Windows:

Enroll an identity into a scratch dir so run-host has something to load:
`.\ziti-edge-tunnel.exe enroll -j C:\temp\id.jwt -i C:\temp\zet-crash\crash.json`

In one terminal:
`.\ziti-edge-tunnel.exe run-host -I C:\temp\zet-crash -P crash`

In another terminal:
`.\ziti-edge-tunnel.exe update_tun_ip -p 16 -P crash`

The command returns `failed to read from pipe: 109`, the first terminal's process is gone, and a minidump is
written to the log folder. Passing a valid `-t` with the same `-p` works and the service stays up.

CLBRITTON2:
- The update_tun_ip handler copied the tun IP before checking one was given, so a command with a prefix and no IP crashed ZET. Move the null check ahead of the copy so it returns "Tun IP is null" (closes #1436)
- Add integration tests for the update_tun_ip rejection paths: missing IP, out-of-range prefix, and malformed IP


tests/integration/testutil/ipc_types.go
```diff
@@ -58,9 +58,9 @@ type MFAData struct {
 }
 
 type TunIPv4Data struct {
-	TunIPv4         string `json:"TunIPv4"`
-	TunPrefixLength int    `json:"TunPrefixLength"`
-	AddDns          bool   `json:"AddDns"`
+	TunIPv4         *string `json:"TunIPv4,omitempty"`
```
dovholuknf, line 61: makes me wnat to pass other junk / nothing on all the other tests too fwiw...

programs/ziti-edge-tunnel/process_cmd.c
```diff
@@ -449,6 +449,12 @@ static void process_update_tun_ipv4_cmd(const char *cmd_json, tunnel_result *res
         result->success = false;
         return;
     }
+    if (tunnel_tun_ip_v4_cmd.tunIP == NULL) {
+        result->error = "Tun IP is null";
```
dovholuknf: "Tun IP cannot be empty" maybe? also did you test "" and " " etc (nothing but whitespace) i didn't see that in the tests?

dovholuknf: and - actually do we validate the ip is parsable as an ip somewhere?

CLBRITTON2: Yeah, the loop right below parses it as a dotted quad, four segments each 0-255, anything else is "Invalid IP address".

dovholuknf (approved):

PR #1438, opened 2026-09-18
implement host.v1 health checks

scareything:


smilindave26 (approved):

programs/ziti-edge-tunnel/ziti-edge-tunnel.c
```diff
@@ -962,6 +962,48 @@ static void on_event(const base_event *ev) {
...
+                        thc->Type = r->check_type ? strdup(r->check_type) : NULL;
+                        thc->IsPassing = r->is_passing;
+                        thc->ConsecutiveFailures = r->consecutive_failures;
+                        thc->Error = r->error ? strdup(r->error) : NULL;
+                        thse.Checks[chk_idx] = thc;
+                    }
+                }
+                send_events_message(&thse, (to_json_fn) tunnel_health_status_event_to_json, true);
+                if (thse.Checks != NULL) {
+                    for (int chk_idx = 0; thse.Checks[chk_idx] != NULL; chk_idx++) {
```
dovholuknf: i think this can be "model-free'ed' as we discussed?

dovholuknf (approved):

tests/integration/health_check_test.go
```diff
@@ -0,0 +1,284 @@
...
+// anyTerminator matches the first terminator seen for a service, regardless of state --
+// used to wait for the initial bind, before any check result has had a chance to move
+// cost/precedence away from baseline.
+func anyTerminator(testutil.Terminator) bool { return true }
+
+// TestHealthCheckPortCheckFailAndRecover hosts a TCP service with a portCheck against
+// its own backend. Taking the backend down should mark the terminator's precedence
+// "failed"; bringing it back should return it to "default".
+func TestHealthCheckPortCheckFailAndRecover(t *testing.T) {
+	testutil.RunWithTimeoutOf(t, 70*time.Second, func(t *testing.T) {
```
CLBRITTON2: After looking at these tests again idk how necessary it really is to use `testutil.RunWithTimeoutOf` while also passing context to `state.overlay.WaitForTerminator`. Perhaps it's a good failover blanket, but in most of the other tests we wait for *some* event or state indefinitely and wrap the entire test in a timeout like this so we either "do the thing" or timeout and fail without having a lot of ctx scattered in the test. Just a thought.

PR #1439, opened 2026-09-21
Reject setting an unknown log level instead of reporting success

CLBRITTON2:
- An unknown or numeric label was reported as success while the tunnel-sdk log level silently dropped to NONE
- Reject unknown, empty, whitespace and numeric labels with an error naming the accepted labels
- Integration tests for each rejection


dovholuknf (approved):
looks good mostly. few comments in dm

